PR #1, opened 2019-11-11
refactor: Port to Application Platform

amcgee:
This should be function, but requires some testing.


amcgee:
Thanks @Sharmyn28, I'll merge this then !

PR #9, opened 2020-02-14
Fix/roots

Sharmyn28:
Hi @amcgee ! In this PR I'm changing:
- the BrowserRouter to HashRouter
- changing two menu section roots: `android-settings` to `general-settings` and `test-android-login` to `test-android-sync`

Let me know if there is something to change or add

### DIFF
--- a/src/components/dialog-table.js
+++ b/src/components/dialog-table.js
@@ -29,7 +29,7 @@ class DialogTable extends React.Component {
                 maxWidth="lg"
             >
                 <DialogTitle id="form-dialog-title">
-                    Values per {this.props.title}
+                    {i18n.t('Values per')} {this.props.title}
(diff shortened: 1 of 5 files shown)


Sharmyn28:
@amcgee here I updated the `en.pot` as we talked, also added some translations.

src/components/dialog-table.js
```diff
@@ -29,7 +29,7 @@ class DialogTable extends React.Component {
                 maxWidth="lg"
             >
                 <DialogTitle id="form-dialog-title">
-                    Values per {this.props.title}
```
amcgee: @Sharmyn28 this should be `i18n.t('values per {{propName}}', { propName: this.props.title })` so that it's easier to translate

PR #10, opened 2020-02-14
Fix/duplicated states

Sharmyn28:
Hi @amcgee ! Here I fixed the duplicated states (specificSettings.name and specificSettingName ...) related to programs and dataSet specific settings, sending `specificSettings` states to `global-specific-setting` component.

Let me know if there something I should add or remove

### DIFF
--- a/src/components/android-settings-container.js
+++ b/src/components/android-settings-container.js
@@ -25,7 +25,7 @@ class AndroidSettingsContainer extends React.Component {
     state = {
         metadataSync: metadataSync,
         dataSync: dataSync,
-        numberSmsToSent: '',
+        numberSmsToSend: '',
         numberSmsConfirmation: '',
         valuesTEI: '',
         encryptDB: encryptDB,
@@ -57,10 +57,10 @@ class AndroidSettingsContainer extends React.Component {
      */
     checkMatchingConfirmation = () => {
         if (
-            this.state.numberSmsToSent !== '' &&
+            this.state.numberSmsToSend !== '' &&
             this.state.numberSmsConfirmation !== ''
         ) {
-            this.state.numberSmsToSent !== this.state.numberSmsConfirmation
+            this.state.numberSmsToSend !== this.state.numberSmsConfirmation
                 ? this.setState({ errorConfirmation: true })
                 : this.setState({ errorConfirmation: false })
         }
@@ -74,8 +74,15 @@ class AndroidSettingsContainer extends React.Component {
             return true
         }
 
-        const androidData = this.state
-        androidData.lastUpdated = new Date().toJSON()
+        const androidData = {
+            metadataSync: this.state.metadataSync,
+            dataSync: this.state.dataSync,
+            numberSmsToSend: this.state.numberSmsToSend,
+            numberSmsConfirmation: this.state.numberSmsConfirmation,
+            valuesTEI: this.state.valuesTEI,
+            encryptDB: this.state.encryptDB,
+            lastUpdated: new Date().toJSON(),
+        }
 
         this.saveDataApi(androidData)
     }
@@ -101,12 +108,11 @@ class AndroidSettingsContainer extends React.Component {
     /**
      * Resets values to default
      */
-    handleReset = e => {
-        e.preventDefault()
+    handleReset = () => {
         this.setState({
             metadataSync: metadataSync,
             dataSync: dataSync,
-            numberSmsToSent: '',
+            numberSmsToSend: '',
             numberSmsConfirmation: '',
             valuesTEI: '',
             encryptDB: encryptDB,
@@ -154,7 +160,7 @@ class AndroidSettingsContainer extends React.Component {
                                       {
                                           metadataSync: metadataSync,
                                           dataSync: dataSync,
-                                          numberSmsToSent: '',
+                                          numberSmsToSend: '',
                                           numberSmsConfirmation: '',
                                           valuesTEI: '',
                                           encryptDB: encryptDB,
@@ -162,11 +168,11 @@ class AndroidSettingsContainer extends React.Component {
                                   )
                                   .then(res => {
                                       this.setState({
-                                          isUpdated: true,
-                                          loading: false,
+                                          /* isUpdated: true,
+                                          loading: false, */
                                           metadataSync: metadataSync,
                                           dataSync: dataSync,
-                                          numberSmsToSent: '',
+                                          numberSmsToSend: '',
                                           numberSmsConfirmation: '',
                                           valuesTEI: '',
                                           encryptDB: encryptDB,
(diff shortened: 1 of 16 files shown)


Sharmyn28:
also in this PR I changed the constants for dataSets and programs (option labels, default values, max values for input number)

src/components/android-settings-container.js
```diff
@@ -74,8 +74,15 @@ class AndroidSettingsContainer extends React.Component {
             return true
         }
 
-        const androidData = this.state
-        androidData.lastUpdated = new Date().toJSON()
+        const androidData = {
```
amcgee: If this is adopting all of the properties of `state` and adding `lastUpdated`, you can use the [object rest spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax):

```js
const androidData = {
  ...this.state,
  lastUpdated: new Date().toJSON()
}
```

This makes a **shallow** clone of the object and adds the `lastUpdated` property - note that if any of the properties are themselves objects,  it will **not** make a copy of them.

src/components/android-settings-container.js
```diff
@@ -154,20 +160,19 @@ class AndroidSettingsContainer extends React.Component {
...
                                           valuesTEI: '',
                                           encryptDB: encryptDB,
                                       }
                                   )
                                   .then(res => {
-                                      console.log(res)
                                       this.setState({
-                                          isUpdated: true,
-                                          loading: false,
+                                          /* isUpdated: true,
```
amcgee: What is the reason for commenting this out?

src/components/android-settings-container.js
amcgee: I see that maybe you want to exclude some properties?  In that case the way you've done it (or making `androidData` a child property of `state`) would be fine

src/components/android-settings-container.js
```diff
@@ -211,7 +215,7 @@ class AndroidSettingsContainer extends React.Component {
             .catch(e => {
                 this.setState({
                     isUpdated: false,
-                    loading: false,
+                    //loading: false,
```
amcgee: Should this be removed instead of commented out?

src/components/dataSet-settings.js
```diff
@@ -59,11 +59,15 @@ class DataSetSettings extends React.Component {
             this.dataSetToChange = args[0].name
             const argsData = args[0]
             this.setState({
-                specificPeriodDSDownload: argsData.specificPeriodDSDownload,
-                specificPeriodDSDBTrimming: argsData.specificPeriodDSDBTrimming,
-                specificSettingName: argsData.id,
+                specificSetting: {
+                    specificSettingName: argsData.id,
```
amcgee: It seems a bit strange to call this `specifiicSettingName` when it is under the `specificSetting` object, should this just be `name`?

src/components/dataSet-settings.js
```diff
@@ -182,42 +198,51 @@ class DataSetSettings extends React.Component {
...
-            specificPeriodDSDBTrimming: this.state.specificPeriodDSDBTrimming,
+            periodDSDownload: this.state.specificSetting.periodDSDownload,
+            periodDSDBTrimming: this.state.specificSetting.periodDSDBTrimming,
         }
 
+        const sumaryString = this.state.specificSetting.periodDSDownload
         const sumarySettings =
-            this.state.specificPeriodDSDownload === undefined
+            this.state.specificSetting.periodDSDownload === undefined
                 ? 0
```
amcgee: Should this really be an integer if undefined and a string if defined?

src/components/dataSet-settings.js
```diff
@@ -409,7 +437,7 @@ class DataSetSettings extends React.Component {
                     api.createNamespace(
                         'ANDROID_SETTING_APP',
                         'dataSet_settings'
-                    ).catch(e => console.log(e))
+                    ).catch(e => console.error(e))
```
amcgee: I see this pattern being used a lot - `.catch(e => console.error(e))`.  It's good to catch the error so it doesn't bubble up to the top level, but this also means that things will just "not work" without any indication to the user that something went wrong.  A good pattern, in general, is to display a user-friendly error when something goes wrong.  This doesn't need to be fixed in this PR, but is something to think about for a future enhancement

Sharmyn28:
Hi @amcgee should I start checking these comments now or wait until we merge the other PR?

amcgee:
You can check them now, unless they are already covered by the other PR

Sharmyn28:
Hi @amcgee! I cleaned the comments that were left, changed settingsName to name and kept the same type of variable. Hope you can check it

src/components/dataSet-settings.js
```diff
@@ -182,72 +198,83 @@ class DataSetSettings extends React.Component {
...
+            }
 
-        if (this.dataSetToChange !== undefined) {
-            let newRowList = []
-            const rowList = this.specificSettingsRows
-            newRowList = rowList.filter(row => row.id !== newDataSetRow.id)
-            newRowList.push(newDataSetRow)
-            this.specificSettingsRows = newRowList
+            if (this.dataSetToChange !== undefined) {
+                let newRowList = []
```
amcgee: This is a minor change, but this appears to be more complicated than it needs to be.  I think it could look like this:

```js
this.specificSettingsRows = this.specificSettingsRows.filter(row => row.id !== newDataSetRow.id)
this.specificSettingsRows.push(newDataSetRow)
```

src/constants/test-android.js
```diff
@@ -107,7 +107,7 @@ export const testAndroidDataConstants = [
         load: 'metadataLoad',
     },
     {
-        title: 'Data download size',
+        title: 'Data download size   (KB)',
```
amcgee: why are there multiple spaces here?  This should also be translated, but that could be a separate PR

src/pages/global-specific-settings.js
```diff
@@ -108,15 +108,15 @@ class GlobalSpecificSettings extends React.Component {
                         handleClose={this.props.specificSettingDialogClose}
                         dataTitle={this.props.specificSettingDataTitle}
                         dataTitleOptions={this.props.specificSettingOptions}
-                        titleValue={this.props.states.specificSettingName}
+                        titleValue={this.props.statesSpecific.name}
```
amcgee: Is this is a typo?  Why is it called `statesSpecific` here?

src/constants/test-android.js
Sharmyn28: I'm cleaning the extra spaces, the translation will be in other PR

src/pages/global-specific-settings.js
Sharmyn28: I'm changing the property name `statesSpecific` to `SpecificSetting` in all components.

src/components/dataSet-settings.js
Sharmyn28: I applied this change also to programs specificSettings

Sharmyn28:
@amcgee I just finished with the changes from the comments we were talking

PR #11, opened 2020-02-25
Fix/translations

Sharmyn28:
adding extra translations for constants like:

- options: programs and datasets
- titles and descriptions: test android
- components with text: test android


Sharmyn28:
@amcgee I'm including translations that were left

PR #12, opened 2020-02-25
Fix/add switch

Sharmyn28:
Module: General Settings
Property: Encrypt DB

- changing radio buttons to a switch
- changing value type from a string ('yes' / 'no') to a boolean

### DIFF
--- a/src/components/android-settings-container.js
+++ b/src/components/android-settings-container.js
@@ -46,9 +46,10 @@ class AndroidSettingsContainer extends React.Component {
                 : (e.target.value = valueInput)
         }
         let value = e.target.value
-        e.target.name === 'encryptDB'
-            ? (value = e.target.checked)
-            : (value = e.target.value)
+
+        if (e.target.name === 'encryptDB') {
+            value = e.target.checked
+        }
 
         this.setState({
             ...this.state,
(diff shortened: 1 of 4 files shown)


Sharmyn28:
@amcgee, I'm adding some changes we were talking yesterday

src/components/android-settings-container.js
```diff
@@ -45,9 +45,14 @@ class AndroidSettingsContainer extends React.Component {
                 ? (e.target.value = maxValues.valuesTEI)
                 : (e.target.value = valueInput)
         }
+        let value = e.target.value
```
amcgee: This should be `const` since it does not change

src/components/android-settings-container.js
```diff
@@ -45,9 +45,14 @@ class AndroidSettingsContainer extends React.Component {
...
+        e.target.name === 'encryptDB'
+            ? (value = e.target.checked)
+            : (value = e.target.value)
```
amcgee: These three lines would be cleaner as a simple `if` statement:

```js
if (e.target.name === 'encryptDB') {
    value = e.target.checked
}
```

Sharmyn28: okey dokey, I just corrected those lines and also regenerated `en.pot`

PR #13, opened 2020-02-25
Fix/style table values tei

Sharmyn28:
- General Settings: removing property `values downloaded per TEI`
- Data set and Program: table text-align.

### DIFF
--- a/src/components/android-settings-container.js
+++ b/src/components/android-settings-container.js
@@ -83,13 +83,13 @@ class AndroidSettingsContainer extends React.Component {
             return true
         }
 
-        let { numberSmsToSend, numberSmsConfirmation } = this.state
-        numberSmsConfirmation == ''
-            ? (numberSmsConfirmation = null)
-            : (numberSmsConfirmation = numberSmsConfirmation)
-        numberSmsToSend == ''
-            ? (numberSmsToSend = null)
-            : (numberSmsToSend = numberSmsToSend)
+        if (this.state.numberSmsToSend === '') {
+            this.state.numberSmsToSend = null
+        }
+
+        if (this.state.numberSmsConfirmation === '') {
+            this.state.numberSmsConfirmation = null
+        }
 
         const androidData = {
             metadataSync: this.state.metadataSync,
--- a/src/components/android-settings-container.js
+++ b/src/components/android-settings-container.js
@@ -7,7 +7,6 @@ import {
     metadataOptions,
     dataOptions,
     androidSettingsDefault,
-    maxValues,
 } from '../constants/android-settings'
 import AndroidSettings from './android-settings'
 
@@ -27,7 +26,6 @@ class AndroidSettingsContainer extends React.Component {
         dataSync: dataSync,
         numberSmsToSend: '',
         numberSmsConfirmation: '',
-        valuesTEI: '',
         encryptDB: encryptDB,
         isUpdated: false,
         loading: true,
@@ -39,12 +37,6 @@ class AndroidSettingsContainer extends React.Component {
      */
     handleChange = e => {
         e.preventDefault()
-        if (e.target.name === 'valuesTEI') {
-            const valueInput = e.target.value
-            e.target.value > maxValues.valuesTEI
-                ? (e.target.value = maxValues.valuesTEI)
-                : (e.target.value = valueInput)
-        }
         this.setState({
             ...this.state,
             [e.target.name]: e.target.value,
@@ -79,7 +71,6 @@ class AndroidSettingsContainer extends React.Component {
             dataSync: this.state.dataSync,
             numberSmsToSend: this.state.numberSmsToSend,
             numberSmsConfirmation: this.state.numberSmsConfirmation,
-            valuesTEI: this.state.valuesTEI,
             encryptDB: this.state.encryptDB,
             lastUpdated: new Date().toJSON(),
         }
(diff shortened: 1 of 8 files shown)


Sharmyn28:
@amcgee I'm adding some changes we were talking about yesterday

Sharmyn28:
In the beginning, the idea was to remove Reserved values (`valuesTEI`) from general settings, but this week we decided to remove `teReserved` from specific and global program settings and put again `valuesTEI` in general settings, also instead of naming it `valuesTEI` we changed to `reservedValues`.

src/components/android-settings-container.js
```diff
@@ -39,12 +44,18 @@ class AndroidSettingsContainer extends React.Component {
      */
     handleChange = e => {
         e.preventDefault()
-        if (e.target.name === 'valuesTEI') {
-            const valueInput = e.target.value
-            e.target.value > maxValues.valuesTEI
-                ? (e.target.value = maxValues.valuesTEI)
-                : (e.target.value = valueInput)
+
+        let { value } = e.target
```
amcgee: As in #12 this should be `const`

amcgee: Also this will conflict with the other change (the line is different) so you'll need to update this PR once the other is merged

src/components/android-settings-container.js
```diff
@@ -74,12 +85,20 @@ class AndroidSettingsContainer extends React.Component {
             return true
         }
 
+        let { numberSmsToSend, numberSmsConfirmation } = this.state
```
amcgee: These are assigned from `state`, but when you re-assign them later (`numberSmsConfirmation = null`, for instance) you are NOT updating the value of `this.data.numberSmsConfirmation` but rather only the value of the variable `numberSmsConfirmation`.  Because you only use `this.state.numberSmsConfirmation` later, this assignment *won't actually work*

src/components/android-settings-container.js
```diff
@@ -74,12 +85,20 @@ class AndroidSettingsContainer extends React.Component {
...
+        numberSmsConfirmation == ''
+            ? (numberSmsConfirmation = null)
+            : (numberSmsConfirmation = numberSmsConfirmation)
```
amcgee: This should be written the other way around if using a ternary operator (`a = a === '' ? b : a`), but here the re-assignment only happens in one condition so it's better and easier-to-read like this:

```js
if (this.state.numberSmsConfirmation === '') {
    this.state.numberSmsConfirmation = null
}
```

src/components/android-settings-container.js
```diff
@@ -74,12 +85,20 @@ class AndroidSettingsContainer extends React.Component {
...
+        numberSmsToSend == ''
+            ? (numberSmsToSend = null)
+            : (numberSmsToSend = numberSmsToSend)
```
amcgee: Same as above

src/components/android-settings-container.js
```diff
@@ -74,12 +85,20 @@ class AndroidSettingsContainer extends React.Component {
...
+
         const androidData = {
             metadataSync: this.state.metadataSync,
             dataSync: this.state.dataSync,
             numberSmsToSend: this.state.numberSmsToSend,
             numberSmsConfirmation: this.state.numberSmsConfirmation,
```
amcgee: You see that here you are using `this.state.numberSmsConfirmation`, so lines 92-94 wouldn't actually change the value saved.

src/components/android-settings-container.js
```diff
@@ -39,12 +44,18 @@ class AndroidSettingsContainer extends React.Component {
...
-                ? (e.target.value = maxValues.valuesTEI)
-                : (e.target.value = valueInput)
+
+        let { value } = e.target
+        e.target.name === 'encryptDB'
+            ? (value = e.target.checked)
+            : (value = e.target.value)
+
+        if (e.target.name === 'reservedValues') {
+            value > maxValues.reservedValues
```
amcgee: Won't `value` be a string here?  You call `parseInt` on it below, but here you are comparing it to a number

src/components/android-settings-container.js
```diff
@@ -39,12 +44,18 @@ class AndroidSettingsContainer extends React.Component {
...
+                ? (value = parseInt(maxValues.reservedValues))
```
amcgee: `maxValues.reservedValues` is a number (`50`), so you don't need to call `parseInt` on it

src/components/android-settings-container.js
amcgee: I would recommend rewriting this line so that it is simpler, like so:

```js
value = Math.min(maxValues.reservedValues, parseInt(value))
```

Sharmyn28: I rewrote it as you said using Math.min, also I'm assigning value to states:
```
this.setState({
            ...this.state,
            [e.target.name]: value,
        })
```

src/components/android-settings-container.js
Sharmyn28: Ohh I see, I forgot to use this assignment, it was supposed to be like this:
```
const androidData = {
            metadataSync: this.state.metadataSync,
            dataSync: this.state.dataSync,
            numberSmsToSend: numberSmsToSend,
            numberSmsConfirmation: numberSmsConfirmation,
            reservedValues: this.state.reservedValues,
            encryptDB: this.state.encryptDB,
            lastUpdated: new Date().toJSON(),
        }
```
but I probably removed that change before commit it, sorry I totally understand this was not working. So now I'm not using this: `let { numberSmsToSend, numberSmsConfirmation } = this.state`   instead I'm just re assigning the states

src/components/android-settings-container.js
Sharmyn28: Yuup I see I'm not using all the re-assigning lines, I'm removing `numberSmsConfirmation` and `numberSmsToSend` and only using states

PR #14, opened 2020-02-25
fix: change program setting from event period to event date

Sharmyn28:
Program module, changing setting name and title from `event period` to `event date`.

### DIFF
--- a/src/components/program-settings.js
+++ b/src/components/program-settings.js
@@ -30,8 +30,8 @@ const {
     teReservedDBTrimming,
     eventsDownload,
     eventsDBTrimming,
-    eventPeriodDownload,
-    eventPeriodDBTrimming,
+    eventDateDownload,
+    eventDateDBTrimming,
 } = ProgramSettingsDefault
 
 class ProgramSettings extends React.Component {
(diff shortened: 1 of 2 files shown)


src/components/program-settings.js
```diff
@@ -67,8 +67,8 @@ class ProgramSettings extends React.Component {
         teReservedDBTrimming: teReservedDBTrimming,
         eventsDownload: eventsDownload,
         eventsDBTrimming: eventsDBTrimming,
-        eventPeriodDownload: eventPeriodDownload,
-        eventPeriodDBTrimming: eventPeriodDBTrimming,
+        eventDateDownload: eventDateDownload,
```
amcgee: This works just fine since it's used throughout this file, but for future reference it is possible to shorten this a bit:

```js
state = {
   eventDateDownload: eventDateDownload,
}
```

is the same as 

```js
state = {
    eventDateDownload,
}
```

And that way you don't need to repeat the name twice.

Sharmyn28: okey dokey! I'll keep it in mind for next PRs

PR #15, opened 2020-03-05
Refactor/namespace constants

Sharmyn28:
Here I changed the namespaces and keynames that we use in Datastore, instead of using the DataStore API with strings I changed to constants. Also, changed `android_settings` keynames to `general_settings`.

### DIFF
--- a/src/components/android-settings-container.js
+++ b/src/components/android-settings-container.js
@@ -10,6 +10,7 @@ import {
     maxValues,
 } from '../constants/android-settings'
 import AndroidSettings from './android-settings'
+import { NAMESPACE, GENERAL_SETTINGS } from '../constants/data-store'
 
 const { metadataSync, dataSync, encryptDB } = androidSettingsDefault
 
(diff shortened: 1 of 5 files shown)


src/components/android-settings-container.js
```diff
@@ -174,16 +165,15 @@ class AndroidSettingsContainer extends React.Component {
                                           numberSmsConfirmation: '',
                                           valuesTEI: '',
                                           encryptDB: encryptDB,
+                                          isUpdated: true,
+                                          loading: false,
```
amcgee: This looks like a functional change, should it be in a different PR?

src/components/dataSet-settings.js
```diff
@@ -370,15 +360,12 @@ class DataSetSettings extends React.Component {
...
-                                                const sumaryString =
-                                                    dataSet.periodDSDownload
                                                 const sumarySettings =
                                                     dataSet.periodDSDownload ===
                                                     undefined
                                                         ? undefined
-                                                        : this.arrangeWord(
-                                                              sumaryString
-                                                          )
+                                                        : dataSet.periodDSDownload
```
amcgee: This also looks like a functional change?

src/components/dataSet-settings.js
```diff
@@ -234,11 +227,10 @@ class DataSetSettings extends React.Component {
                     .periodDSDBTrimming,
             }
 
-            const sumaryString = this.state.specificSetting.periodDSDownload
             const sumarySettings =
                 this.state.specificSetting.periodDSDownload === undefined
                     ? undefined
-                    : this.arrangeWord(sumaryString)
+                    : this.state.specificSetting.periodDSDownload
```
amcgee: This also looks like a functional change?

src/components/android-settings-container.js
Sharmyn28: I will remove it and create a new PR just for this change

src/components/dataSet-settings.js
Sharmyn28: Indeed this is a functional change, the thing is that on my dataStore I was already using the input number so it sent me a render error, I think I'm also removing these lines in another PR, I'll keep the lines.

src/components/dataSet-settings.js
Sharmyn28: I'll keep the lines

Sharmyn28:
Hi @amcgee I just removed the functional settings

PR #16, opened 2020-03-05
Fix/enum values program

Sharmyn28:
This change has:
- Remove triple m from `teiDBTrimming` max values constants
- Keep settings with number type as number instead of a string

### DIFF
--- a/src/components/program-settings.js
+++ b/src/components/program-settings.js
@@ -179,24 +179,23 @@ class ProgramSettings extends React.Component {
 
     handleChange = e => {
         e.preventDefault()
-        let { value } = e.target
-        value = this.chooseSetting(e.target.name, value)
         this.setState({
             ...this.state,
-            [e.target.name]: value,
+            [e.target.name]: this.chooseSetting(e.target.name, e.target.value),
         })
         this.updateGlobal = true
     }
 
     handleChangeDialog = e => {
         e.preventDefault()
-        let { value } = e.target
-        value = this.chooseSetting(e.target.name, value)
         this.setState({
             ...this.state,
             specificSetting: {
                 ...this.state.specificSetting,
-                [e.target.name]: value,
+                [e.target.name]: this.chooseSetting(
+                    e.target.name,
+                    e.target.value
+                ),
             },
         })
         this.updateGlobal = false
--- a/src/components/program-settings.js
+++ b/src/components/program-settings.js
@@ -142,34 +142,25 @@ class ProgramSettings extends React.Component {
     chooseSetting = (name, value) => {
         switch (name) {
             case 'teiDownload':
-                value > maxValues.teiDownload
-                    ? (value = parseInt(maxValues.teiDownload))
-                    : (value = parseInt(value))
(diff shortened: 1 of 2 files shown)


src/components/program-settings.js
```diff
@@ -139,94 +139,64 @@ class ProgramSettings extends React.Component {
...
+    }
 
+    handleChange = e => {
+        e.preventDefault()
+        let { value } = e.target
+        value = this.chooseSetting(e.target.name, value)
         this.setState({
             ...this.state,
-            [e.target.name]: e.target.value,
+            [e.target.name]: value,
```
amcgee: I don't think it's necessary to use a mutable variable here, you could do this instead:

```js
        const value = this.chooseSetting(e.target.name, e.target.value)
        this.setState({
            ...this.state,
            [e.target.name]: value,
         })
```

(you could also just call `this.chooseSetting` directly on line 186, but the temporary `const` variable is fine too)

src/components/program-settings.js
```diff
@@ -139,94 +139,64 @@ class ProgramSettings extends React.Component {
...
+                    ? (value = parseInt(maxValues.eventsDownload))
+                    : (value = parseInt(value))
                 break
             case 'eventsDBTrimming':
-                e.target.value > maxValues.eventsDBTrimming
-                    ? (e.target.value = maxValues.eventsDBTrimming)
-                    : (e.target.value = valueInput)
+                value > maxValues.eventsDBTrimming
+                    ? (value = parseInt(maxValues.eventsDBTrimming))
+                    : (value = parseInt(value))
```
amcgee: All of these would be MUCH simpler written as `value = Math.min(maxValues.eventDBTrimming, parseInt(value))`, as I mentioned in another review.  The `parseInt` on `maxValues.x` isn't necessary, since they should already be numbers, and this way the assignment only happens once.  Would you mind rewriting each of these cases that way?

src/components/program-settings.js
```diff
@@ -139,94 +139,64 @@ class ProgramSettings extends React.Component {
...
-                e.target.value > maxValues.eventsDBTrimming
-                    ? (e.target.value = maxValues.eventsDBTrimming)
-                    : (e.target.value = valueInput)
-                break
-            default:
-                break
-        }
-
+        let { value } = e.target
+        value = this.chooseSetting(e.target.name, value)
```
amcgee: Similarly, this doesn't need to be a mutable variable

src/components/program-settings.js
Sharmyn28: I'm just calling `this.chooseSetting` directly on line 184

src/components/program-settings.js
Sharmyn28: I'm removing the mutable variable and calling `this.chooseSetting` directly on line 194

src/components/program-settings.js
Sharmyn28: Sure I rewrite every case like this:
`value = Math.min(maxValues.teiDownload, parseInt(value))`

PR #17, opened 2020-03-05
Fix/period option data set

Sharmyn28:
This PR has:

- change Period Option from a Select Option to an Input Number.
- Options constants were removed

### DIFF
--- a/src/components/dataSet-settings.js
+++ b/src/components/dataSet-settings.js
@@ -227,12 +227,6 @@ class DataSetSettings extends React.Component {
         this.updateGlobal = false
     }
 
-    arrangeWord = word => {
-        let newString = word.toUpperCase().split('_')
-        newString = newString.join(' ')
-        return newString
-    }
-
     /**
      * Submit Dialog data, specific settings
      */
@@ -254,7 +248,6 @@ class DataSetSettings extends React.Component {
                     .periodDSDBTrimming,
             }
 
-            //const sumaryString = this.state.specificSetting.periodDSDownload
             const sumarySettings =
                 this.state.specificSetting.periodDSDownload === undefined
                     ? undefined
--- a/src/components/dataSet-settings.js
+++ b/src/components/dataSet-settings.js
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
         },
     }
 
+    chooseSetting = (name, value) => {
+        switch (name) {
+            case 'periodDSDownload':
+                value = parseInt(value)
+                break
+            case 'periodDSDBTrimming':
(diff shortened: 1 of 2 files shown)


src/components/dataSet-settings.js
```diff
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
         },
     }
 
+    chooseSetting = (name, value) => {
+        switch (name) {
+            case 'periodDSDownload':
+                value = parseInt(value)
```
amcgee: Rather than assigning to the `value` argument, why not just `return parseInt(value)` here?  That way you also don't need the `break` line

src/components/dataSet-settings.js
```diff
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
...
+                break
+            case 'periodDSDBTrimming':
+                value = parseInt(value)
```
amcgee: This could also be just `return parseInt(value)`

src/components/dataSet-settings.js
```diff
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
...
+                break
+            default:
+                break
```
amcgee: this could be `return value`

src/components/dataSet-settings.js
```diff
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
...
+        }
+        return value
+    }
+
     handleChange = e => {
         e.preventDefault()
+        let { value } = e.target
```
amcgee: This doesn't need to be mutable (same as #16)

src/components/dataSet-settings.js
```diff
@@ -83,22 +83,42 @@ class DataSetSettings extends React.Component {
...
             ...this.state,
-            [e.target.name]: e.target.value,
+            [e.target.name]: value,
         })
         this.updateGlobal = true
     }
 
     handleChangeDialog = e => {
         e.preventDefault()
+        let { value } = e.target
```
amcgee: This doesn't need to be mutable (same as #16)

src/components/dataSet-settings.js
```diff
@@ -234,11 +254,11 @@ class DataSetSettings extends React.Component {
                     .periodDSDBTrimming,
             }
 
-            const sumaryString = this.state.specificSetting.periodDSDownload
+            //const sumaryString = this.state.specificSetting.periodDSDownload
```
amcgee: If this is no longer needed let's remove it instead of commenting it out!

PR #19, opened 2020-03-09
Fix/default values2

Sharmyn28:
This PR has:

-  **General Settings**: In case there is no value for `numberSmsToSend` and `numberSmsConfirmation` should send null also as default value instead of an empty string.
- **DataSet Settings**: In case a dataset is selected but there is no setting should send as a default value instead of an empty string.
- **Program Settings**: specific settings should have default values


src/components/program-settings.js
```diff
@@ -69,20 +70,22 @@ class ProgramSettings extends React.Component {
...
-            enrollmentDBTrimming: '',
-            enrollmentDateDownload: '',
-            enrollmentDateDBTrimming: '',
-            updateDownload: '',
-            updateDBTrimming: '',
-            eventsDownload: '',
-            eventsDBTrimming: '',
-            eventDateDownload: '',
-            eventDateDBTrimming: '',
+            settingDownload: SpecificSettingsDefault.settingDownload,
```
jenniferarnesen: Looks like you should be able to just spread SpecificSettingsDefault here instead of setting each prop individually.

```specificSetting: {
openDialog: false,
name: '',
...SpecificSettingsDefault
}```

File: src/constants/program-settings.js [hunk 2]
```diff
@@ -415,3 +415,20 @@ export const ProgramSettingsDefault = {
     eventDateDownload: 'ANY',
     eventDateDBTrimming: 'ANY',
 }
+
+export const SpecificSettingsDefault = {
+    settingDownload: 'ALL_ORG_UNITS',
```
jenniferarnesen: Perhaps 'ALL_ORG_UNITS' and 'ANY' should be defined as constants?

src/components/android-settings-container.js
```diff
@@ -31,8 +33,8 @@ class AndroidSettingsContainer extends React.Component {
     state = {
         metadataSync: metadataSync,
         dataSync: dataSync,
-        numberSmsToSend: '',
-        numberSmsConfirmation: '',
+        numberSmsToSend: numberSmsToSend,
```
jenniferarnesen: You can use object shorthand in a lot of places in this file (probably elsewhere too)

PR #20, opened 2020-03-10
Feature/program type

Sharmyn28:
This PR has:

- Before show specificSettings options user should select a program, then the app will check programType property:

1. `WITH_REGISTRATION`: Should show only these options (settingDownload, settingDBTrimming, teiDownload, teiDBTrimming, enrollmentDownload, enrollmentDBTrimming, enrollmentDateDownload, enrollmentDateDBTrimming, updateDownload, updateDBTrimming) 

2. `WITHOUT_REGISTRATION`: Should show only (settingDownload, settingDBTrimming, eventsDownload, eventDateDownload, eventDateDownload, eventDateDBTrimming)

- Changes in program_settings constants (specificSettings, withRegistration, withoutRegistration)  SpecificSettingsDefault constants are not included because prev PR (defaultValues) includes these constants.
- SettingsTable (component) is replacing ProgramTable (component), ProgramTable render a table using programType property


src/components/dialog-table.js
```diff
@@ -9,73 +9,92 @@ import MenuItem from '@material-ui/core/MenuItem'
...
+    return (
+        <Dialog
+            open={open}
+            onClose={handleClose}
+            aria-labelledby="form-dialog-title"
+            fullWidth
+            maxWidth="lg"
+        >
+            <DialogTitle id="form-dialog-title">
+                {i18n.t('Values per')} {title}
```
jenniferarnesen: You should be able to use i18n interpolation here

Issue #7, opened 2020-01-13
Sidebar is incorrect when refreshing page


amcgee:
The sidebar doesn't take the current route into account when refreshing

PR #28, opened 2020-04-20
Fix/menu side bar

Sharmyn28:
This PR fixes Issue #7 Sidebar, when refreshing takes the current route

### DIFF
--- a/src/pages/layout.js
+++ b/src/pages/layout.js
@@ -24,29 +24,23 @@ const Layout = () => {
                 <SideBar />
                 <MainContent>
                     <Paper className={layoutStyles.paper__layout}>
-                        <D2Shim>
-                            <Switch>
+                        <Switch>
+                            <Route
+                                path="/"
+                                exact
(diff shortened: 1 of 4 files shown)


src/pages/layout.js
```diff
@@ -1,70 +1,42 @@
...
-                    </header>
                     <Paper className={layoutStyles.paper__layout}>
                         <D2Shim>
                             <Switch>
-                                {routeSections.map(section => (
+                                <Route
+                                    path="/"
+                                    exact
+                                    render={() => (
+                                        <D2Shim>
```
jenniferarnesen, line 33: Curious if it is necessary to have another D2Shim when there is already one above on line 27?

Sharmyn28: I tried to remove D2Shim from lines 33 and 44 and keep only the one from line 27 but it fails to get users and the amount of data that the user would sync.

I can remove it from line 27 and 33, and keep the one inside of the menuSection map (line 44).

PR #30, opened 2020-04-23
Feat: First launch dialog

Sharmyn28:
This PR has:

First launch where a configuration has not yet been saved. The user must take explicit action to save the default configuration to get started. The first-launch state also provides an opportunity to clearly state the consequences of applying settings in this app.

In this PR I'm having troubles to handle states, I want to set state after all promises are resolved so the dialog closes and the general section is shown. I hope you can give me an idea of how to solve it.

### DIFF
--- a/src/pages/layout.js
+++ b/src/pages/layout.js
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React from 'react'
 
 import { TwoPanel, Sidebar, MainContent } from '@dhis2/d2-ui-core'
 import { Paper } from '@material-ui/core'
@@ -10,6 +10,17 @@ import i18n from '@dhis2/d2-i18n'
 
 import { D2Shim } from '../utils/D2Shim'
 import layoutStyles from '../styles/Layout.module.css'
+import api from '../utils/api'
+import {
+    DATASET_SETTINGS,
+    GENERAL_SETTINGS,
+    NAMESPACE,
+    PROGRAM_SETTINGS,
+} from '../constants/data-store'
+import { androidSettingsDefault } from '../constants/android-settings'
+import { SpecificSettingsDefault } from '../constants/program-settings'
+import { DataSetSettingsDefault } from '../constants/data-set-settings'
+import DialogFirstLaunch from '../components/dialog/dialog-first-launch'
 
 const styles = {
     twoPanelMain: {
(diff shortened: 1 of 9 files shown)


jenniferarnesen:
> In this PR I'm having troubles to handle states, I want to set state after all promises are resolved so the dialog closes and the general section is shown. I hope you can give me an idea of how to solve it.

It should be possible to do this by putting the setState into a then():
```
api.createNamespace()
.then(Promise.all[...])
.then(() => {
   this.setState(...)
});
```

jenniferarnesen:
There is a conflict between the FirstLaunch dialog and the General settings page:

The FirstLaunch modal displays because the api has not found the ANDROID namespace. But at the same time, the General Settings page is loaded initially (that is the default route), which triggers `createNamespace` in componentDidMount. 

So then, when I click the button in FirstLaunch to create the defaults, I get an error that the namespace already exists (which is now true because General Settings page has just created it).

The UX/Product person might want to be involved to help decide on the best workflow here. Should the general settings page even be loaded if there is no namespace? Instead of a modal, should there be a First Launch "landing page"?

Related to this, when I click Cancel in the modal, I am redirected to the dhis2 dashboards, which kind of surprised me. I'm thinking that instead of "Cancel", it should say where the user is going to be redirected to (ie., "Return to dhis2 home page"). Maybe it isn't even a button, but an inconspicuous link at the bottom. This might also be a question for the UX or Product manager.

jenniferarnesen:
Also, you are missing a `return` in the createNamespace function that is keeping it from working. This line:

https://github.com/dhis2/android-settings-app/blob/feat/firstTimeLaunch/src/utils/api.js#L32

Because namespace.set returns a promise, and the namespace isn't actually created until this Promise is resolved. So the function could look like this:

```
createNamespace(namespace, key, value = {}) {
    return getInstance()
      .then((d2) =>
        d2.dataStore.create(namespace).then((namespace) => {
          return namespace.set(key, value);
        })
      )
      .catch((error) => Promise.reject(error));
  }
```

src/pages/layout.js
```diff
@@ -10,76 +10,189 @@ import i18n from '@dhis2/d2-i18n'
...
-
-    const changeSectionHandler = key => {
-        currentSection.current = key
-        if (key !== 'search' && sidebarRef.current) {
-            sidebarRef.current.clearSearchBox()
-        }
+const DEFAULT_PROGRAM = 'DEFAULT_PROGRAM'
+const DEFAULT_DATASET = 'DEFAULT_DATASET'
+
+const populateObject = (type, settingsList) => {
```
jenniferarnesen, line 34: How about instead of passing `settingsList`, you just use the correct object (either DataSetSettingsDefault or SpecificSettingsDefault) based on the `type`?

src/pages/layout.js
```diff
@@ -10,76 +10,189 @@ import i18n from '@dhis2/d2-i18n'
...
+    let object
```
jenniferarnesen, line 35: Set this initially to an empty object, so that the function doesn't return undefined in the default case.

src/pages/layout.js
```diff
@@ -10,76 +10,189 @@ import i18n from '@dhis2/d2-i18n'
...
+import api from '../utils/api'
+import {
+    DATASET_SETTINGS,
+    GENERAL_SETTINGS,
+    NAMESPACE,
+    PROGRAM_SETTINGS,
+} from '../constants/data-store'
+import { androidSettingsDefault } from '../constants/android-settings'
+import { SpecificSettingsDefault } from '../constants/program-settings'
+import { DataSetSettingsDefault } from '../constants/data-set-settings'
```
jenniferarnesen, line 22: Could you make these objects lowerCase, like `androidSettingsDefault`? This is for convention and consistency :)

Sharmyn28:
> There is a conflict between the FirstLaunch modal and the General settings page:
> 
> The FirstLaunch modal displays because the api has not found the ANDROID namespace. But at the same time, the General Settings page is loaded initially (that is the default route), which triggers `createNamespace` in componentDidMount.
> 
> So then, when I click the button in FirstLaunch to create the defaults, I get an error that the namespace already exists (which is now true because General Settings page has just created it).
> 
> The UX/Product person might want to be involved to help decide on the best workflow here. Should the general settings page even be loaded if there is no namespace? Instead of a modal, should there be a First Launch "landing page"?
> 
> Related to this, when I click Cancel in the modal, I am redirected to the dhis2 dashboards, which kind of surprised me. I'm thinking that instead of "Cancel", it should say where the user is going to be redirected to (ie., "Return to dhis2 home page"). Maybe it isn't even a button, but an inconspicuous link at the bottom. This might also be a question for the UX or Product manager.

Hi Jennifer! The `FirstLaunch` modal should be displayed when there is no ANDROID_SETTINGS_APP  namespace, so the idea is that when clicking on the button it should create a namespace with 3 keys (general_settings, dataSet_settings, and program_settings) and their default values. The only place where we should create namespace is on this dialog, the pages or sections should only update data. This is a new function.

After the user clicks on the button and all the values have been created, the General Settings should be displayed with the default data, in case there are a namespace and a key it should show the data from datastore. 

Click on Cancel is a still in progress idea, and it was a possibility that it should redirect to a default link.  

In #29 I’m removing the `createNamespace` from general, dataset, and program settings, this change is not part of that PR because I was having trouble with the state and I wanted if it was possible that you can lend me a hand to check this.

src/pages/layout.js
Sharmyn28: Seems like a great idea, I just solved it

src/pages/layout.js
Sharmyn28: okey dokey, also applying this

src/pages/layout.js
Sharmyn28: hehehe okey dokey, thanks for the advice :)

Sharmyn28:
> > In this PR I'm having troubles to handle states, I want to set state after all promises are resolved so the dialog closes and the general section is shown. I hope you can give me an idea of how to solve it.
> 
> It should be possible to do this by putting the setState into a then():
> 
> ```
> api.createNamespace()
> .then(Promise.all[...])
> .then(() => {
>    this.setState(...)
> });
> ```

I tried to apply this, adding the return to `createNamespace` function, and removing the `createNamespace` from General settings but not sure why it's not working as I wish

src/utils/api.js
```diff
@@ -24,17 +24,13 @@ class Api {
         )
     }
 
-    createNamespace(namespace, key, value) {
+    createNamespace(namespace, key, value = {}) {
         return getInstance()
             .then(d2 => {
-                console.log(d2.dataStore)
                 d2.dataStore.create(namespace).then(namespace => {
```
jenniferarnesen, line 30: I think you also need a `return` on this line. e.g.

return d2.dataStore.create(...

Sharmyn28: hehe completely forgot about this, thanks now it's working, I'll save it and it will be ready

PR #39, opened 2020-05-13
Feat/remove btn

Sharmyn28:
This PR has a suggestion for disabling the settings/configuration using a button.

Workflow:
- You find the button only in General section
- Clicking on the button will remove/delete the namespace and all the keys.
- After that will show you the first launch dialog.


src/components/dialog/dialog-first-launch.js
```diff
@@ -6,66 +6,64 @@ import {
...
+                    </p>
 
-                        <strong>
-                            {i18n.t(
-                                'Any settings or configuration on a users device will be overwritten by settings applied here.'
-                            )}
-                        </strong>
+                    <strong>
+                        {i18n.t(
+                            'Any settings or configuration on a users device will be overwritten by settings applied here.'
```
jenniferarnesen, line 36: user's

src/pages/layout.js
```diff
@@ -96,48 +93,42 @@ class Layout extends React.Component {
...
                     <MainContent>
                         <Paper className={layoutStyles.paper__layout}>
                             <Switch>
-                                <Route
-                                    path="/"
-                                    exact
-                                    render={() => (
+                                <Route path="/" exact>
+                                    {this.state.openFirstLaunch === true ? (
+                                        <D2Shim>
```
jenniferarnesen, line 118: Is it possible to have the <D2Shim> component wrap the ternary statement all together?

```
<D2Shim>
{this.state.openFirstLaunch === true ? (...) : (...)}</D2Shim>

src/components/dialog/dialog-first-launch.js
Sharmyn28: thanks jennifer :)

src/pages/layout.js
Sharmyn28: Just checked this option and it's working

PR #40, opened 2020-05-14
Fix/confirm sms field

Sharmyn28:
This PR has:

- Change label of `number SMS confirmation`, use the same label use in the android app. This was talked to the android team.
- Add validation to gateway and confirmation number using the same regular expression from the app.
- Fix null input console warning, by not sending empty values, null or undefined to datastore. This solves this [Jira issue](https://jira.dhis2.org/browse/DHIS2-8801). Also, this was talked to the SDK team and confirmed that not sending gateway or confirmation number to datastore will not generate errors with SDK.


jenniferarnesen:
Looks like even if I enter an invalid value, I can still save. If there is going to be phone number validation, I think the feedback should have more information ("Incorrect phone number" doesn't really help).

I would suggest adding some unit tests for the validateNumber function, if you have time.

image

Sharmyn28:
Hi, @jenniferarnesen if I add this text to validate and disable the `save` button it will be a better validation?

image

Sharmyn28:
Changing the helper text and error message to:

image

jenniferarnesen:
It seems I need to add 4 characters before the number is considered valid:

Sharmyn28:
Hi @jenniferarnesen this is the regEx and function to validate that I'm using:

```
const validateNumber = number => {
    const phoneRegex = RegExp('^\\+[1-9][0-9]{3,16}\$')
    return phoneRegex.test(number)
}
```

So when I try to validate this is what happens:

```
validateNumber("+1234")
**true**
```

```
validateNumber("+123")
**false**
```

so better to put 4 characters long?

jenniferarnesen:
> so better to put 4 characters long?

Yes, that matches better with the behavior.

PR #41, opened 2020-05-14
Fix: specific settings empty value

Sharmyn28:
This PR has:

- Prevent app crashes with an empty value (Data set specific settings) by checking if empty value change it to 0, solving [Jira 8795](https://jira.dhis2.org/browse/DHIS2-8795), [Jira 8861](https://jira.dhis2.org/browse/DHIS2-8861) and [Jira 8862](https://jira.dhis2.org/browse/DHIS2-8862)
- Fix console warning: input value should not be null or NaN, solving [Jira 8808](https://jira.dhis2.org/browse/DHIS2-8801)
- Add a label with the Default value for numeric properties (dataset and program, specific settings)
- Fix different libraries used for table children (table body, table row, table cell) in the` program table-row` component.
- Fix, when change program check that settings are set to default values.


src/components/settings-table/dataset-table-row.js
```diff
@@ -19,7 +21,12 @@ const DataSetTableRow = ({
                 scope="row"
                 className={dataTableStyles.dataTable__rows__row__column}
             >
-                {dataRow.option} ({periodType})
+                <p>
+                    {dataRow.option} ({periodType})
+                </p>
+                <em className={tableTitleStyles.attributeLabel}>
+                    {i18n.t('Default:')} {defaultValue}
```
jenniferarnesen, line 28: Suggest using interpolation for this translation:

`i18n.t('Default: {{defaultValue}}', {defaultValue})`

src/modules/prepareSpecificSettingToSave.js
```diff
@@ -71,14 +68,10 @@ export const prepareSpecificSettingsToSave = ({
                     )
                 }
 
-                summarySettings =
-                    (states.specificSetting.eventsDownload
-                        ? states.specificSetting.eventsDownload
-                        : specificSettingsDefault.eventsDownload) +
-                    ' events per OU'
+                summarySettings = `${states.specificSetting.eventsDownload} events per OU`
```
jenniferarnesen, line 71: Do these need to be translated with i18n?

src/components/settings-table/dataset-table-row.js
Sharmyn28: as you suggest I'm separating this phrase in 3 parts:
`{i18n.t('Default')} : {i18n.t('{{defaultValue}}', {defaultValue})}`

src/modules/prepareSpecificSettingToSave.js
Sharmyn28: I'm translating this line as well as line 53 and 83

src/components/settings-table/dataset-table-row.js
```diff
@@ -25,7 +25,8 @@ const DataSetTableRow = ({
                     {dataRow.option} ({periodType})
                 </p>
                 <em className={tableTitleStyles.attributeLabel}>
-                    {i18n.t('Default:')} {defaultValue}
+                    {i18n.t('Default')} :{' '}
+                    {i18n.t('{{defaultValue}}', { defaultValue })}
```
jenniferarnesen, line 29: After seeing what ends up in en.pot, I think this way of doing it will be really hard for the translators to figure out. Do you get the error if you escape the separator?  e.g. `\:`. If not, maybe a better way of doing this would be:
`i18n.t('Default{{separator}} {{defaultValue}}', {separator: ":", defaultValue})`

Sharmyn28: I get the error if I use the separator. I'll try it using the other option

PR #42, opened 2020-05-14
fix: data download loading, max value states

Sharmyn28:
This PR has:

- fix data load state, so it can properly show value of data download size
- fix translation for `Recommended maximum:`
- change maxValue states from `undefined` to 0


Sharmyn28:
Hi @jenniferarnesen I was checking user test and the flag that shows the loader or the data download size was only showing the loader even when the data size was ready to be shown

image

PR #44, opened 2020-05-14
feat: user filter with query option

Sharmyn28:
This PR has:

- Change user filter options, from `username` and `name` only to all options by using a query.

Solving [Jira 8796](https://jira.dhis2.org/browse/DHIS2-8796)


amcgee:
Generally looks good!

Suggested to @Sharmyn28 on a call : 
1. Move `onChange` function to a variable, so it doesn't need to be defined in-line
2. See if it's possible to use `<D2Shim>` instead of passing the d2 prop through multiple layers

src/components/text-field-search.js
```diff
@@ -94,6 +96,32 @@ export default class TextFieldSearch extends React.Component {
...
     }
 
+    handleFilter = (event, clearSelection) => {
+        this.props.clearFields()
+        if (event.target.value === '') {
+            clearSelection()
+            this.suggestions = []
+            this.setState({
+                suggestionsList: [],
+            })
```
jenniferarnesen, line 106: Are you sure `clearSelection()` is needed? I removed it and the list still cleared as expected. Also, what are the different purposes of `suggestions` and `suggestionsList` ?

Sharmyn28: Hi Jennifer! `clearSelection()` is a downshift action that clears the selection by cleaning some internal states like: 

```
selectedItem: null,
inputValue: ''
```

And suggestions and suggestionsList are basically the same but before the last change this.state.suggestionsList was not updating as it was supposed, I check it again and it's working so I'm fixing the duplicated state/property.

PR #46, opened 2020-05-26
Fix: hash history and reorder disable settings

Sharmyn28:
This PR has:

- Add history from createHashHistory and change from hashRouter to Router, to access navigation history.
- Add a fallback route to catch no match routes. `Page not found `component, in case the URL is not part of the routes.
- Change `Disable all settings` button position, add dialog to ensure knowledge of "Disable" action
- Fix reserved values constants

Fixing bug:
The fix/implement is related to a bug found on the "Disable settings" workflow. When clicking on disable settings the namespace and keys should be deleted from datastore and after that because there is no namespace the workflow starts again from first setup dialog. 
Bug:  When the user clicked on Disable Settings, they were afterward redirected to a different page  where the app is not located


jenniferarnesen:
Notes from Milagros regarding the reasons for the switch away from HashRouter:

**************************
 I tried to upload the app to a server to test the "disable settings" btn (after removing the namespace the method used to replace the current path to "/" so if there is no namespace it renders the first launch compent), and on my local it was working ok using the hashRouter.
But when I tested on a server if the path was for example: https://debug.dhis2.org/dev/api/apps/Android-Settings/index.html#/general-setting
After replacing the pathname to "/", the path will end up like this: https://debug.dhis2.org/dev
This will not render the first launch dialog and neither the app because is not the right location.
The pathname that I need it as a replacing result was: https://debug.dhis2.org/dev/api/apps/Android-Settings/index.html#/
I checked other options and I found history, to access the location, pathname and history. I tried to add the history to the routes to pass it as a props to the section components (general, program and dataset settings) but it didn't work because the structure is:
<D2Shim> {component} </D2Shim>
so I found that instead of creating just a browserHistory I could also create a hashHistory. hashRouter doesn't allow history as a prop but Router component can pass history to children, so I at the end this is why I changed it. Using Router but also hashHistory, it solved the problem with the paths.

jenniferarnesen:
> I checked other options and I found `history`, to access the location, pathname and history. I tried to add the history to the routes to pass it as a props to the section components (general, program and dataset settings) but it didn't work because the structure is:
> `<D2Shim> {component} </D2Shim>`

Did you try the `useHistory` hook? https://reacttraining.com/react-router/web/api/Hooks/usehistory

src/modules/general/apiLoadSettings.js
```diff
@@ -16,6 +17,8 @@ export const apiLoadGeneralSettings = async () => {
                         })
                 }
             })
+        } else {
+            navigateTo('/')
```
jenniferarnesen, line 21: I think it would be better to have this happening in a component, rather than here in the api function.

PR #47, opened 2020-06-15
feat: specify datastore keys permission

Sharmyn28:
This PR has:

- Sharing POST for datastore keys when setup for the first time.
- Check Authority, only if the user has ALL authority settings should be enabled to create, edit, or delete settings. If not all settings and buttons should be disabled, read-only.
- If the user has ALL authority can create the namespace in datastore otherwise this option should be disabled
- Only show program/data set with specific settings that are related to the current user


src/components/settings-table/setting-table-row.js
```diff
@@ -6,11 +6,18 @@ import { TableRow, TableCell, RadioGroup, Radio } from '@dhis2/ui-core'
...
-            <TableCell> {dataRow.option} </TableCell>
+            <TableCell
+                className={
+                    states.disableAll ? disable.disable_label : undefined
+                }
+            >
+                {dataRow.option}
+            </TableCell>
             <TableCell
                 className={`${dataTableStyles.dataTable__rows__row__column} ${dataTableStyles.dataTable_align_end}`}
```
amcgee, line 22: This should also use `classnames`

src/components/table-actions.js
```diff
@@ -30,12 +31,18 @@ const TableActions = ({ columns, rows, menuActions }) => {
                 {rows.map(row => (
                     <TableRow key={row.id}>
                         <TableCell
-                            className={dataTableStyles.dataTable_row_title}
+                            className={`${
+                                dataTableStyles.dataTable_row_title
+                            } ${states.disableAll &&
```
amcgee, line 36: This also needs to use `classnames`

src/components/table-actions.js
```diff
@@ -30,12 +31,18 @@ const TableActions = ({ columns, rows, menuActions }) => {
...
+                                disableStyle.disable_label}`}
                         >
                             {row.name}
                         </TableCell>
                         <TableCell
-                            className={dataTableStyles.dataTable_row_title}
+                            className={`${
+                                dataTableStyles.dataTable_row_title
+                            } ${states.disableAll &&
```
amcgee, line 44: And another place

src/modules/general/apiLoadSettings.js
```diff
@@ -1,21 +1,25 @@
...
-                if (res.includes(GENERAL_SETTINGS)) {
-                    return api
-                        .getValue(NAMESPACE, GENERAL_SETTINGS)
-                        .then(res => {
-                            return res.value
-                        })
-                }
-            })
-        }
+    return getAuthority().then(hasAuthority => {
```
amcgee, line 9: It would be better not to n'est a bunch of callbacks here, this is very difficult to understand at a glance.  I would recommend splitting it out into well-named helper functions

src/modules/programs/apiLoadSettings.js
```diff
@@ -16,74 +17,90 @@ export const apiLoadProgramSettings = async ({
...
-                                programListComplete,
-                                specificSettingsRows,
-                                globalSettings,
-                                globalDefaultValues,
-                            }
-                        })
-                }
-            })
-        }
+                                res.value.globalSettings.disableAll = !hasAuthority
```
amcgee, line 89: It's better to treat this as constant, rather than mutating it - so you should create a copy below.

src/components/table-actions.js
Sharmyn28:
```
className={cx(dataTableStyles.dataTable_row_title, {
                                [disableStyle.disable_label]: states.disableAll,
                            })}
```

Sharmyn28:
@amcgee I've just added some changes to fix it.

amcgee:
Milagros, you can merge this but *please create an issue to address the wording change I suggested, as well as the mutation of res.value in apiLoadSettings*

Sharmyn28:
> Milagros, you can merge this but _please create an issue to address the wording change I suggested, as well as the mutation of res.value in apiLoadSettings_

Thanks Austin, I've just created issues for these comments.

PR #54, opened 2020-09-11
fix: [DHIS2-9131] change ui components for data set and programs

Sharmyn28:
**Description**
Change material design components like TextField and Dialog to UI components SingleSelect, InputField, Modal in Dataset, and Program Settings.

[Check sketch](https://www.sketch.com/s/7437a41c-0ddc-44ea-bfac-d75ca6620bae/a/7QjjOY/play)

jira issue [DHIS2-9131](https://jira.dhis2.org/browse/DHIS2-9131), [DHIS2-9132](https://jira.dhis2.org/browse/DHIS2-9132), [DHIS2-9133](https://jira.dhis2.org/browse/DHIS2-9133) and [DHIS2-9134](https://jira.dhis2.org/browse/DHIS2-9134)

**Solution description**

- Add UI
- Use SingleSelect, InputField (number), Table, Modal
- Changing components for global and specific settings (dataset and program)


src/components/inputs/input-number.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { InputField } from '@dhis2/ui'
+
+export const InputNumber = ({ data, states, onChange }) => (
```
amcgee, line 4: This should take `keyDownload`, `maxValue`, `value`, and `disableAll` as props directly, I think.  It also need a PropTypes declaration.

src/components/inputs/select.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const Select = ({
+    onChange,
+    label,
+    name,
+    value,
+    disabled,
+    options,
+    inputWidth,
+}) => (
```
amcgee, line 12: This component looks good, it has all the props explicitly!  It should also have PropTypes, but this is a good model for the other Select components

src/components/sections/dataset/dataSet-settings.js
```diff
@@ -277,17 +271,18 @@ class DataSetSettings extends React.Component {
...
             })
         },
-        onInputChange: e => {
-            e.preventDefault()
+        onInputChange: (e, key) => {
+            let name, value
+
+            typeof key === 'string'
+                ? ((name = key), (value = e.selected))
+                : ((name = e.name), (value = e.value))
```
amcgee, line 279: This is indirect logic, shouldn't use a ternary in this way if you're not assigning the result of the ternary to some variable.

There are two simpler, easier-to-read options:

```
const name = typeof key === 'string' ? key : e.name
const value = typeof key === 'string' ? e.selected : e.value
```

OR (less desirable because it doesn't allow the variables to be const, but a little cleaner because you don't need any ternary):

```
let name, value

if (key === 'string') {
  name = key
  value = e.selected
} else {
  name = e.name
  value = e.value
}
```

src/components/sections/program/program-settings.js
```diff
@@ -138,36 +136,26 @@ class ProgramSettings extends React.Component {
...
-                    e.target.value === PER_ORG_UNIT
-                ) {
-                    programData = GlobalProgramSpecial
-                } else {
-                    programData = GlobalProgram
-                }
-            }
+        typeof key === 'string'
+            ? ((name = key), (value = e.selected))
+            : ((name = e.name), (value = e.value))
```
amcgee, line 144: Again this shouldn't use a ternary in this way, see above.

src/components/sections/program/program-settings.js
```diff
@@ -138,36 +136,26 @@ class ProgramSettings extends React.Component {
...
-                ...this.state,
-                disableSave: false,
-                [e.target.name]: parseValueByType(
-                    e.target.name,
-                    e.target.value
-                ),
-            })
+        if (name === SETTING_DOWNLOAD) {
+            if (value === GLOBAL || value === PER_ORG_UNIT) {
+                programData = GlobalProgramSpecial
```
amcgee, line 148: I know this was here before, but it is **incorrect** - `programData` is defined as a singleton variable for this entire module, meaning there is only ever **one value for the entire app**.  Instead, this needs to be a state variable in the component itself (the state is set just below this).

amcgee: @Sharmyn28 let me know if there's a reason you made this a singleton

src/components/sections/program/program-settings.js
```diff
@@ -344,21 +331,21 @@ class ProgramSettings extends React.Component {
...
-                        ),
+                        [key]: parseValueByType(key, e.selected),
                     },
                 })
             } else {
+                let name, value
+
+                typeof key === 'string'
+                    ? ((name = key), (value = e.selected))
+                    : ((name = e.name), (value = e.value))
```
amcgee, line 342: See above, same issue

src/components/sections/section-wrapper.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { CircularLoader } from '@dhis2/ui'
+import UnsavedChangesAlert from '../unsaved-changes-alert'
+
+const SectionWrapper = ({ loading, unsavedChanges, ...props }) => {
```
amcgee, line 5: Props only needs to contain `children` here, so `children` should be explicitly listed as a prop.  This also needs a PropTypes declaration

src/components/settings-table/dataset-table.js
```diff
@@ -1,126 +1,65 @@
...
-                dataSet => dataSet.id === this.props.dataSetSelected
+    const checkPeriodType = edit => {
+        if (props.dataSetSelected !== '') {
+            const dataSetSelected = props.completeListOptions.filter(
+                dataSet => dataSet.id === props.dataSetSelected
             )
-            this.dataSetFilter = dataSetSelected[0]
+            if (!edit) {
+                props.states.periodDSDownload =
+                    periodTypeConstants[dataSetSelected[0].periodType].default
```
amcgee, line 21: This will directly mutate a deeply-nested value in `props` which **will update that value without rerendering any components**.  Mutation should always happen in a function which calls `setState` (perhaps in a component higher up the tree)

src/components/settings-table/dataset-table.js
```diff
@@ -1,126 +1,65 @@
...
-                                onChange={this.props.onChange}
-                            />
-                        ))}
-                    </TableBody>
-                </Table>
-            )
-        }
+    })
 
+    if (!(dataSetSelected && periodType)) {
```
amcgee, line 44: This would be a little clearer as `if (!dataSetSelelcted || !periodType)`

src/components/settings-table/program-table-row.js
```diff
@@ -1,26 +1,18 @@
...
-    return (
+const ProgramTableRow = ({ dataRow, states, onChange }) => (
+    <div>
         <TableRow>
-            <TableCell
-                component="th"
-                scope="row"
-                className={dataTableStyles.dataTable__rows__row__column}
-            >
+            <div>
```
amcgee, line 15: Why is this no longer a `TableCell`?

src/components/settings-table/program-table.js
```diff
@@ -1,106 +1,63 @@
...
+    const [programFilter, setProgramFiltered] = useState('')
+    const [programType, setProgramType] = useState('')
 
-    checkProgramType = () => {
-        if (this.props.programSelected !== '') {
-            const programSelected = this.props.completeListOptions.filter(
-                program => program.id === this.props.programSelected
+    const checkProgramType = () => {
+        if (props.programSelected !== '') {
+            const programSelected = props.completeListOptions.filter(
```
amcgee, line 14: This overrides a variable name higher up, use a different name

src/components/wrapper.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import cx from 'classnames'
+import styles from '../styles/TableActions.module.css'
+
+const Wrapper = props => (
```
amcgee, line 5: This should take `children` directly as a prop, and needs `propTypes`

src/modules/dataset/parseValueBySettingType.js
```diff
@@ -7,7 +7,7 @@ export const parseValueBySettingType = (name, value) => {
     switch (name) {
         case PERIOD_DS_DOWNLOAD:
         case PERIOD_DS_DB_TRIMMING:
-            return value ? parseInt(value) : 0
+            return value <= 0 ? 0 : parseInt(value)
```
amcgee, line 10: `value` is a string here right?  Why would it ever be `<= 0`?

amcgee (changes requested):
Added some comments, let me know if you want to discuss!

It would be great to make these PRs much smaller - i.e. just do this for one component so that we can review them quicker and learn from the experience of each small change, rather than needing to make lots of small corrections in a very big change

src/components/settings-table/program-table-row.js
Sharmyn28: This is part of Joe's suggestions [(sketch)](https://www.sketch.com/s/7437a41c-0ddc-44ea-bfac-d75ca6620bae/a/7QjjOY/play)

src/components/settings-table/dataset-table.js
Sharmyn28: I've moved this mutation to `dataSet-settings` thanks for the advice.

src/modules/dataset/parseValueBySettingType.js
Sharmyn28: Hi @amcgee! yup `value` is a string, but when using the arrows from number input you can get a **negative integer**.  Period values should only be positive integers.

I added a console.log using `return value ? parseInt(value) : 0`  and this was the result  `{name: "periodDSDownload", value: "-1"}`

amcgee: I think this will work, oddly, because `"-1" <= 0` in weird Javascript logic.  However, I think it would be a lot clearer to do this:

```js
return Math.max(0, parseInt(value))
```

This assumes that `value` will always be a valid positive or negative number, not an arbitrary string.  If it could be an arbitrary string it would need to be:

```js
isNaN(value) ? 0 : Math.max(0, parseInt(value))
```

Screen Shot 2020-11-02 at 4 01 39 PM

Sharmyn28: I was checking this for an arbitrary string:

image

the thing is that when is an empty string the value is NaN, if there is no value it should be 0; so I was thinking in do it like this:

`isNaN(parseInt(value)) ? 0 : Math.max(0, parseInt(value)) `
 
image

src/components/settings-table/dataset-table-row.js
```diff
@@ -1,46 +1,39 @@
...
 const DataSetTableRow = ({
     dataRow,
     periodType,
     defaultValue,
     states,
     onChange,
-}) => {
-    return (
+}) => (
+    <div>
```
amcgee, line 16: It's better to use `<>` or `<React.Fragment>` here, instead of rendering an extra Div unnecessarily

src/components/settings-table/dataset-table.js
```diff
@@ -1,126 +1,59 @@
...
-        if (this.props.dataSetSelected !== '') {
-            const dataSetSelected = this.props.completeListOptions.filter(
-                dataSet => dataSet.id === this.props.dataSetSelected
+const DataSetTable = props => {
+    const [dataSet, setDataset] = useState(props.dataSetSelected)
+    const [dataSetData, setData] = useState(props.data)
+    const [periodType, setPeriodType] = useState('')
+    const [defaultValue, setDefault] = useState('')
+    const [dataSetFilter, setDataSetFiltered] = useState('')
+    const [selected, setSelected] = useState(false)
```
amcgee, line 12: It's better to use a single state object (`useState({ dataSet, periodType, ...etc })`) than a bunch of useStates when the state pieces are related

src/components/settings-table/program-table.js
```diff
@@ -1,106 +1,59 @@
...
-        if (this.props.programSelected !== '') {
-            const programSelected = this.props.completeListOptions.filter(
-                program => program.id === this.props.programSelected
+import Wrapper from '../wrapper'
+
+const ProgramTable = props => {
+    const [program, setProgram] = useState(props.programSelected)
+    const [programData, setData] = useState(props.data)
+    const [programFilter, setProgramFiltered] = useState('')
+    const [programType, setProgramType] = useState('')
```
amcgee, line 10: Again, this could be a bit cleaner with a single `useState` object

src/modules/programs/parseValueBySettingType.js
```diff
@@ -9,22 +9,28 @@ import {
 export const parseValueByType = (name, value) => {
     switch (name) {
         case TEI_DOWNLOAD:
-            value = value ? Math.min(maxValues.teiDownload, parseInt(value)) : 0
+            value =
+                isNaN(parseInt(value)) || Math.max(0, parseInt(value)) <= 0
```
amcgee, line 13: This is a bit too complicated... `Math.max(0, parseInt(value))` will never be `< 0`, could be cleaned up:

```
Math.min(maxValues.teiDownload, Math.max(0, isNan(value) ? 0 : parseInt(value)))
```

It would also be cleaner to pull this out into a helper function, since it is repeated multiple places.

amcgee (approved):
Phew that was a big one!  Let's get this in - I made a few smaller comments, but feel free to merge this and fix them in separate (smaller!) PRs

amcgee:
Nice work @Sharmyn28 🎉

PR #55, opened 2020-09-12
feat: [DHIS2-9156] Add TEI reserved value to user sync test

Sharmyn28:
**Description**
Add the number of TEI reserved value to user sync test

jira issue: [DHIS2-9156](https://jira.dhis2.org/browse/DHIS2-9156)

**Solution description**   
- Get value from datastore (general_settings, teiReservedValue)
- Show value and recommended value

### DIFF
--- a/src/components/sections/user-sync-test/run-test.js
+++ b/src/components/sections/user-sync-test/run-test.js
@@ -45,8 +45,13 @@ const RunTest = ({ states }) => {
                                         {test.title}
                                     </p>
                                     <p className={itemStyles.subItemItem}>
-                                        {i18n.t('Recommended maximum:')}
-                                        {test.maxValue}
+                                        {i18n.t(
+                                            'Recommended maximum: {{maxValue}}',
+                                            {
+                                                nsSeparator: '---',
+                                                maxValue: test.maxValue,
+                                            }
+                                        )}
                                     </p>
                                 </div>
                             </Grid>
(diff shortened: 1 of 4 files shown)


src/components/sections/user-sync-test/run-test.js
```diff
@@ -44,7 +45,7 @@ const RunTest = ({ states }) => {
                                         {test.title}
                                     </p>
                                     <p className={itemStyles.subItemItem}>
-                                        {i18n.t('Recommended maximum:')}{' '}
+                                        {i18n.t('Recommended maximum:')}
```
amcgee, line 48: This should use an i18next interpolation variable (`i18n.t('Recommended maximum: {{maxValue}}', { maxValue: test.maxValue })`)

amcgee (approved):
This logic is very complex and these functions should be simplified, but the change looks OK (made one minor suggestion)

src/components/sections/user-sync-test/run-test.js
Sharmyn28: Hi Austin! When I use interpolation I get this:

image

Sharmyn28:
> This logic is very complex and these functions should be simplified, but the change looks OK (made one minor suggestion)

Great! I'm working on refactoring the code but as a part of another PR :)

src/components/sections/user-sync-test/run-test.js
amcgee: Sorry @Sharmyn28, my fault!  This is due to the `:` character being used as the namespace separator.  You can avoid this with (I believe):

```js
i18n.t('Recommended maximum: {{maxValue}}', { nsSeparator: '---', maxValue: maxValue })
```

You can confirm it works by making sure `Recommended maximum: {{maxValue}}` shows up in the `en.pot` file

Sharmyn28: Thanks, this did the trick, I checked the `en.pot `

PR #56, opened 2020-10-09
fix: [DHIS2-9301] add Manual option to sync

Sharmyn28:
Add manual option to Sync Metadata and Data

- when clicking on the manual option, an alert dialog should pop up
- if click on OK, the manual option will be selected
- if click on Cancel, the previous option will remain selected
- This option only available from android app version 2.3.0 onwards

**Alert Dialog Text:** 
```
By selecting Manual there will NOT be any AUTOMATIC SYNCHRONIZATION of data. Your users will need to sync manually for data to be sent to the server. Do you want to select this option?'
```

[JIRA DHIS2 9301](https://jira.dhis2.org/browse/DHIS2-9301)

### DIFF
--- a/src/constants/android-settings.js
+++ b/src/constants/android-settings.js
@@ -9,6 +9,10 @@ export const metadataOptions = [
         value: '7d',
         label: '7d',
     },
+    {
+        value: 'manual',
+        label: 'Manual',
+    },
 ]
 
 export const dataOptions = [
@@ -31,6 +35,10 @@ export const dataOptions = [
     {
         value: '24h',
         label: '24h',
+    },
+    {
+        value: 'manual',
+        label: 'Manual',
     },
 ]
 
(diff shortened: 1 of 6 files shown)


src/constants/android-settings.js
```diff
@@ -51,6 +59,7 @@ export const androidSettingsDefault = {
 }
 
 export const RESERVED_VALUES = 'reservedValues'
+export const manual = 'manual'
```
HendrikThePendric, line 62: It'd be more consistent to uppercase this variable name. Also you probably want to refer to this constant instead of using a string literal in lines 13 and 40 above.

HendrikThePendric (approved):
Code is looking good and the functionality is working as expected. Great job 🎉 .

I've left a minor comment. It's be great if you could address this, but since it's only a minor cosmetic thing, I've approved already.

src/constants/android-settings.js
HendrikThePendric: (this would also mean moving all the constant declarations up in the file, which would make sense to me as well)

Sharmyn28: Thanks, Hendrik! I will fix it :)

PR #62, opened 2021-02-25
fix: get namespace and authority using app runtime

Sharmyn28:
- Migrate d2 API to app runtime, get namespace and authority
- Add new keys constants for namespace


src/components/router.js
```diff
@@ -9,24 +9,20 @@ import DialogFirstLaunch from './dialog/dialog-first-launch'
 const Router = () => {
     const [openFirstLaunch, setFirstLaunch] = useState(true)
     const [isSaved, setSaved] = useState(false)
```
jenniferarnesen, line 11: Looks like `isSaved` is not in use anywhere, so should be safe to remove it.

src/modules/apiLoadFirstSetup.js
```diff
@@ -1,15 +1,35 @@
...
+
+export const namespaceDataStoreQuery = {
+    dataStore: {
+        resource: 'dataStore',
+    },
+}
+
+export const authorityQuery = {
+    authority: {
+        resource: 'me/authorization/ALL',
```
jenniferarnesen, line 12: Should this be `authorization` or `authorities`?  Probably best to ask someone on the api team. I just wondered because I was checking out the new documentation and saw `authorities` documented. Seems they both work, but one may be preferred over the other.

src/components/router.js
```diff
@@ -9,24 +9,20 @@ import DialogFirstLaunch from './dialog/dialog-first-launch'
...
+        if (namespace && authority) {
+            setFirstLaunch(false)
+            setSaved(true)
+            setAuthority(!authority)
+        } else if (namespace === false && authority !== undefined) {
+            setFirstLaunch(true)
+            setSaved(false)
+            setAuthority(!authority)
+        }
+    }, [namespace, authority])
```
jenniferarnesen, line 25: I'm not sure I totally understand what's going on. But it seems like this component could just use the `namespace` value, and the DialogFirstLaunch component could use useDataQuery on its own to get the authorities and set the disabled value.

Sharmyn28: Oh, ok, I'll try that. In case later, I would like to share to other components if a user has `Authority`, should I use a Context?

src/modules/apiLoadFirstSetup.js
Sharmyn28: I've just asked, it seems both resources are identical. and `authorities` is preferable as it's more descriptive of what it is, but are identical. Should I change it to `authorities`?

jenniferarnesen (approved):

PR #63, opened 2021-02-26
feat: [DHIS2-9585] General Settings section

Sharmyn28:
General settings form [Jira](https://jira.dhis2.org/browse/DHIS2-10403)

- Create basic fields form (text, number, phone number, URL, checkbox)
- Create a basic layout page with a header (title and description)
- Get and save settings in the data store using app runtime mutation ([JIRA](https://jira.dhis2.org/browse/DHIS2-10593)).
- Remove namespace from the data store using app runtime mutation.
- Add key names for data store (constants)

image


i18n/en.pot
```diff
@@ -117,16 +117,34 @@ msgstr ""
...
+"Encrypt all data stored on device. Data can be lost if there are problems "
+"with an encrypted database. This will not affect the DHIS2 database stored "
+"on an external server."
 msgstr ""
 
-msgid "No match for"
+msgid "Analytics reporting ID (Matomo)"
+msgstr ""
+
+msgid "Must be at least 1 characters long"
```
jenniferarnesen, line 132: characters -> character

src/components/field/EncryptDB.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from 'react'
+import PropTypes from '@dhis2/prop-types'
+import i18n from '@dhis2/d2-i18n'
+import { CheckboxField } from './CheckboxField'
+import DialogEncrypt from '../dialog/dialog-encrypt'
+
+const CODE = 'encryptDB'
+const LABEL = i18n.t('Encrypt device database')
+const HELPTEXT = i18n.t(
+    'Encrypt all data stored on device. Data can be lost if there are problems with an encrypted database. This will not affect the DHIS2 database stored on an external server.'
```
jenniferarnesen, line 10: I'm not sure if this is still the case, but we've had problems with strings not translating when they are defined outside of a function (@amcgee has this been solved? - I'm thinking yes, but want to check)

src/utils/validators/objectsAreEqual.js
```diff
@@ -0,0 +1,16 @@
+export const objectsAreEqual = (object1, object2) => {
```
jenniferarnesen, line 1: What about using lodash `isEqual` instead?

src/utils/validators/isValidValue.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * A valid value it is consider if it is:
+ * value == true
+ * length > 0
+ * */
+
+export const isValidValue = value =>
```
jenniferarnesen, line 7: Add unit tests for this and `isValidUrl`

src/pages/General/queryMutation.js
```diff
@@ -0,0 +1,14 @@
+import { GENERAL_SETTINGS, NAMESPACE } from '../../constants/data-store'
```
jenniferarnesen, line 1: The filename `queryMutation` is perhaps a little too generic?

src/modules/useDatastore.js
```diff
@@ -0,0 +1,21 @@
...
+ * Query to get General Settings from Data store
+ * */
+
+export const getGeneralKeyQuery = {
+    generalSettings: {
+        resource: `dataStore/${NAMESPACE}/${GENERAL_SETTINGS}`,
+    },
+}
+
+export const useGeneralDataStore = () => {
```
jenniferarnesen, line 14: Would it work to just use `useDataQuery` directly in the component, rather than wrapping it in another hook?

src/components/field/EncryptDB.js
amcgee: This should be OK in platform applications which fetch and load the relevant locale before initializing the application code, but it is safer to ensure that `i18n.t` is always called at render time, rather than at module load time.  This way, if the React tree is re-rendered after switching locales it will correctly pick up the new language.

src/utils/validators/objectsAreEqual.js
Sharmyn28: Good idea!

src/utils/validators/isValidValue.js
Sharmyn28: I added some test like these:

```
test('undefined is not valid value', () => {
    expect(isValidValue(undefined)).toBeFalsy()
})
```

```
test('"https://debug.dhis2.org" is a valid URL', () => {
    const url = 'https://debug.dhis2.org'
    expect(isValidURL(url)).toBeTruthy()
})
```

jenniferarnesen: That's a good start. Could be expanded later.

src/pages/General/GeneralSettings.js
```diff
@@ -0,0 +1,109 @@
...
+    return (
+        <Page
+            title={i18n.t('General Settings')}
+            loading={loading}
+            unsavedChanges={!disableSave}
+        >
+            {settings && (
+                <>
+                    <MatomoUrl value={settings} onChange={setSettings} />
+
```
jenniferarnesen, line 82: Suggest removing these empty lines in the jsx.

jenniferarnesen (approved):

PR #67, opened 2021-04-22
feat: [DHIS2-10404] Create Global Synchronization section

Sharmyn28:
Synchronization: Global Section ([DHIS2-10404](https://jira.dhis2.org/browse/DHIS2-10404))

- Queries to read and update Synchronization key from the data store, using app runtime mutation
- Select fields for data and metadata sync

image


src/components/field/MetadataSync.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import i18n from '@dhis2/d2-i18n'
+import PropTypes from '@dhis2/prop-types'
+import { MANUAL, SyncSelect } from './SyncSelect'
+
+const CODE = 'metadataSync'
+const metadataOptions = [
+    {
+        value: '24h',
```
jenniferarnesen, line 9: Line 22 has a constant defined with this value. Does it make sense to use the constant here instead?

jenniferarnesen (approved):

PR #79, opened 2021-05-13
feat: [DHIS2-11077] Synchronization Dataset specific setting, new, edit and delete

Sharmyn28:
Synchronization: Dataset specific settings (DHIS2-11077)

- Show specific settings list on an action table with a summary of their settings
- Add specific setting:
    - datasets with different period type have a different default value
- Edit specific setting
- Delete specific setting

**Table Action**
image

**New Specific settings**
image
image
image

**Edit specific settings**
image

**Delete specific settings**
image


i18n/en.pot
```diff
@@ -624,6 +624,26 @@ msgstr ""
 msgid "General Settings"
 msgstr ""
 
+msgid "Data sets specific download sync settings"
+msgstr ""
+
+msgid ""
+"Applies only to the assigned data set. Data set specific settings will "
```
jenniferarnesen, line 631: Seems to be some inconsistency in use of "data set" and "dataset". Could be a good idea to agree on the terminology with the product manager.

src/utils/api.js
```diff
@@ -55,33 +55,9 @@ class Api {
...
-                return ret
-            })
-        }
-
-        return new Promise(resolve => {
-            console.log('cache resolve')
-            resolve(cache[namespace][key])
+        return getInstance().then(d2 => {
+            const url = `dataStore/${namespace}/${key}`
+            return d2.Api.getApi().get(url)
```
jenniferarnesen, line 60: Will this be replaced with dataQuery?

jenniferarnesen (approved):

i18n/en.pot
Sharmyn28: Changing "dataset" to "data set"

PR #81, opened 2021-07-01
refactor: delete unused components

Sharmyn28:
The current PR has:

- remove components that are not use: 
    - android/general sync settings
    - program sync settings
    - data sets sync settings
    - svg icons
- reorganize files and directories

Note: the logic behind the code has not been changed or modified.


mediremi (approved):
LGTM :heavy_check_mark: 

Only change to make that I can see is that `src/components/field/Fied.module.css` should be `src/components/field/Field.module.css`

PR #91, opened 2021-11-05
refactor: [appearance] colocate data with component table

Sharmyn28:
The current PR colocate data with the component table (Home and Data set)


mediremi (approved):

src/pages/Appearance/Datasets/TableSettings.js
```diff
@@ -1,11 +1,42 @@
...
 import PropTypes from '@dhis2/prop-types'
+import i18n from '@dhis2/d2-i18n'
 import { TableRowWrapper } from '../../../components/table'
 import { CheckboxField } from '../../../components/field'
-import {
-    datasetAppearanceSettings,
-    datasetAppearanceSpecificSettings,
-} from '../../../constants/dataset-appearance'
+
+const datasetAppearanceSettings = [
```
mediremi, line 7: If the amount of settings on different pages with the same keys grows in the future, translations could be moved to one file called something like `settingsDisplayNames`, which would contain an object like this:

```js
export const settingsDisplayNames = {
  period: i18n.t('Period'),
  organisationUnit: i18n.t('Organisation Unit'),
  // ...
}
```

This file could then be imported by the various settings pages and be used like so:

```js
// TableSettings
import settingsDisplayNames from 'settingsDisplayNames.js'
const datasetAppearanceSettings = ['period', 'organisationUnit', ...]

// In the DatasetGlobalSettings component
datasetAppearanceSettings.map(key => {
  const displayName = settingsDisplayNames[key]
  // ...
})
```

This would keep translations in sync and consistent.

mediremi (approved):

PR #93, opened 2021-11-06
refactor: [analytics] use common query and mutation

Sharmyn28:
Locate GET and UPDATE Analytics Datastore in a common file for TEI Analytics and Visualizations


src/pages/Analytics/analyticsDatastoreQuery.js
```diff
@@ -32,15 +33,38 @@ export const getAnalyticsKeyQuery = {
...
 export const useReadAnalyticsDataStore = () => {
-    const { loading, error, data } = useDataQuery(getAnalyticsKeyQuery)
+    const [teiAnalytics, setTei] = useState()
+    const [visualizations, setVisualizations] = useState()
+    const [home, setHome] = useState()
+    const [program, setProgram] = useState()
+    const [dataSet, setDataSet] = useState()
+
+    const { loading, error } = useDataQuery(getAnalyticsKeyQuery, {
+        onComplete: result => {
```
mediremi, line 43: Because of [the way app-runtime v3 caches queries](https://developers.dhis2.org/blog/2021/11/app-platform-v8#new-feature-usedataquery-caches-and-deduplicates-queries), using onComplete in this way would make it hard to migrate this app to app platform v8/app runtime v3 in the future.

Two ways to refactor this would be:
- Revert to previous method of returning relevant fields from `data`
- Move code inside `onComplete` to a `useEffect` hook with `data` as its dependency

src/pages/Analytics/TEI/AnalyticsTEI.js
```diff
@@ -28,12 +33,11 @@ const AnalyticsTEI = () => {
     }, [hasAuthority])
 
     useEffect(() => {
-        if (analytics) {
-            const settings = analytics.analytics.tei
-            setInitialValues(settings)
-            setAnalyticSettings(settings)
+        if (tei) {
+            setInitialValues(tei)
```
mediremi, line 37: Is `initialValues` still required or can `tei` be used directly instead? It seems like `initialValues` and `tei` are kept in sync?

mediremi (changes requested):

src/pages/Analytics/analyticsDatastoreQuery.js
Sharmyn28: I'm refactoring it like this: 
```
useEffect(() => {
        if (data) {
            const { tei, dhisVisualizations } = data.analytics
            setTei(tei || [])
            setHome(dhisVisualizations ? dhisVisualizations.home : [])
            setProgram(dhisVisualizations ? dhisVisualizations.program : {})
            setDataSet(dhisVisualizations ? dhisVisualizations.dataSet : {})
            setVisualizations(
                dhisVisualizations
                    ? dhisVisualizations
                    : {
                        home: [],
                        program: {},
                        dataSet: {},
                    }
            )
        }
    }, [data])
```

mediremi (approved):

PR #95, opened 2021-11-11
feat: add visualizations search field

Sharmyn28:
This PR adds a visualization search field, that will be used as a common component.

_Visualization search field_
image

image


mediremi (approved):
LGTM :heavy_check_mark:

PR #96, opened 2021-11-12
feat: [DHIS2-11261] Add new visualization for programs

Sharmyn28:
This PR has:

- Basic components such as Text field to add a title or name a visualization and Select program
- Query to generate a random id
- Radio Group: 
  - new group that shows a text field
  - field to select a group that has been created previously
 

_Dialog to add visualization_
image

_Title field_
image

_Group field_
image

image

image

### DIFF
--- a/src/components/analyticVisualization/SelectProgram.js
+++ b/src/components/analyticVisualization/SelectProgram.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import i18n from '@dhis2/d2-i18n'
 import PropTypes from '@dhis2/prop-types'
 import { SelectField } from '../field'
@@ -6,13 +6,7 @@ import { useReadProgramQuery } from '../../pages/Analytics/Program/ProgramVisual
 
 export const SelectProgram = ({ settings, onChange }) => {
     const { programList, loading } = useReadProgramQuery()
-    const [options, setOptions] = useState([])
-
-    useEffect(() => {
-        if (programList) {
-            setOptions(programList)
-        }
-    }, [programList])
+    const options = programList || []
 
     const handleChange = e => {
         const name = options.find(program => program.id === e.selected)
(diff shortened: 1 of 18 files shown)


src/pages/Analytics/Program/helper.js
```diff
@@ -97,3 +125,43 @@ export const createDataStoreGroupRows = datastore => {
...
+    } else {
+        updatedRow[current.program] = {
+            programName: current.programName,
+            groups: {
+                [current.group.id]: [current],
+            },
+        }
+    }
+
+    return updatedRow
```
mediremi, line 166: I haven't tested it, but the following should be equivalent:

```suggestion
    const programRow = rows[current.program]
    
    if (programRow) {
        const group = programRow.groups[current.group.id]
        const updatedGroups = {
            ...programRow.groups,
            [current.group.id]: group ? [...group, current] : [current]
        }

        return {
            ...rows,
            [current.program]: {
                ...programRow,
                groups: updatedGroups
            }
        }
    }
    
    return {
        ...rows,
        [current.program]: {
            programName: current.programName,
            groups: {
                [current.group.id]: [current],
            }
        }
    }
```

src/components/analyticVisualization/SelectProgram.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const SelectProgram = ({ settings, onChange }) => {
+    const { programList, loading } = useReadProgramQuery()
+    const [options, setOptions] = useState([])
+
+    useEffect(() => {
+        if (programList) {
+            setOptions(programList)
+        }
+    }, [programList])
```
mediremi, line 15: I think this should work too?

```suggestion
    const { programList, loading } = useReadProgramQuery()
    const options = programList || []
```

src/pages/Analytics/Program/helper.js
```diff
@@ -1,4 +1,32 @@
...
+    program: initialValues.program || '',
+    visualization: initialValues.visualization || '',
+    name: initialValues.name || '',
+    group: initialValues.group || {
+        name: '',
+        id: '',
+    },
+})
+
+export const validMandatoryFields = settings => {
```
mediremi, line 15: Should this function be called `invalidMandatoryFields` instead? Seems like it's false when `settings.program` and `settings.visualization` are valid

src/pages/Analytics/Program/helper.js
Sharmyn28: Thank you, it's working

mediremi (approved):

PR #105, opened 2021-12-22
fix: [DHIS2-12254] remove hard limits and add warning

Sharmyn28:
This PR has:

- remove limits for Reserved values and add a warning
- remove limits for TEI and Events, add a warning

_Reserved values warning_
image

_TEI warning_ 
image

_Events warning_
image


i18n/en.pot
```diff
@@ -356,6 +356,9 @@ msgstr ""
 msgid "Reserved values downloaded per TEI attribute"
 msgstr ""
 
+msgid "It is recommended a maximum of 500 reserve values"
```
mediremi, line 359: I think this message can be a bit shorter: `Recommended maximum is 500 reserve values`

src/components/inputs/InputNumber.js
```diff
@@ -10,19 +13,39 @@ export const InputNumber = ({
...
+            break
+    }
+
+    const showWarning = value >= maxValue
+    const validationText = showWarning
+        ? i18n.t(
+              'It is recommended to download a maximum of {{maxValue}} {{key}}',
+              { maxValue, key }
+          )
+        : ''
```
mediremi, line 33: `TEI` and `events` may be referred to differently in different languages, so I'd suggest generating the validation text like this:

```suggestion
    let warning = ''
    switch (keyDownload) {
        case 'teiDownload':
            warning = i18n.t('Recommended maximum is {{maxValue}} TEI', { maxValue })
            break
        case 'eventsDownload':
            warning = i18n.t('Recommended maximum is {{maxValue}} events', { maxValue })
            break
    }

    const showWarning = value >= maxValue
    const validationText = showWarning ? warning : ''
```

mediremi (approved):

PR #107, opened 2022-01-11
feat: general section Share screen

Sharmyn28:
This PR has: 

- component to select if a user wants to share screen
- add default share screen value to initial values

image


src/components/field/ShareScreen.js
```diff
@@ -0,0 +1,32 @@
...
+        <CheckboxField
+            name={CODE}
+            label={i18n.t('Take screenshots and share it')}
+            helpText={i18n.t(
+                'Allow user to take screenshots and share their screen.'
+            )}
+            checked={value[CODE]}
+            onChange={handleCheckbox}
+            {...props}
+        />
```
mediremi, line 25: Since the help text is pretty descriptive and short, I think we can use it as the label:

```suggestion
        <CheckboxField
            name={CODE}
            label={i18n.t(
                'Allow user to take screenshots and share their screen.'
            )}
            checked={value[CODE]}
            onChange={handleCheckbox}
            {...props}
        />
```

mediremi (approved):

PR #108, opened 2022-01-11
fix: DHIS2-12338 validate function for url and protocol

Sharmyn28:
This PR has:

valid URL should start with HTTP or https


src/utils/validators/__tests__/isValidURL.test.js
```diff
@@ -5,27 +5,32 @@ test('"https://debug.dhis2.org" is a valid URL', () => {
...
-})
-
 test('"http://www.aa.com" is a valid URL', () => {
     const url = 'http://www.aa.com'
     expect(isValidURL(url)).toBeTruthy()
 })
 
-test('"apple.gr" is a valid URL', () => {
-    const url = 'apple.gr'
+test('"http://matomo.com" is not a valid URL', () => {
```
mediremi, line 13: Typo?
```suggestion
test('"http://matomo.com" is a valid URL', () => {
```

src/utils/validators/__tests__/isValidURL.test.js
```diff
@@ -5,27 +5,32 @@ test('"https://debug.dhis2.org" is a valid URL', () => {
...
     const url = 'http://future'
     expect(isValidURL(url)).toBeFalsy()
 })
+
+test('"htp://future" is not a valid URL', () => {
+    const url = 'htp://future'
+    expect(isValidURL(url)).toBeFalsy()
+})
+
+test('"apple.gr" is a valid URL', () => {
```
mediremi, line 33: Typo?
```suggestion
test('"apple.gr" is not a valid URL', () => {
```

src/utils/validators/isValidURL.js
```diff
@@ -1,6 +1,6 @@
 export const isValidURL = url => {
     const res = url.match(
-        /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g
+        /(https?:\/\/)(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-zA-Z0-9()]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g
```
mediremi, line 3: Since this check is now more strict and requires the https protocol to be present in the url, this function can be replaced by:

```js
export const isValidURL = url => {
  try {
    const url = new URL(string)
    return url.protocol === "https:"
  } catch (error) {
    return false
  }
}
```

Sharmyn28: I like the protocol suggestion, but I was thinking if I should use a regular expression,  a valid URL should be like this:

- must start with either http or https and
- length (2, 256)
- contains domain like .com, .org etc.

mediremi: if more flexibility is needed then yeah a regular expression might make more sense

mediremi (approved):

PR #111, opened 2022-01-26
fix: DHIS2-12339 visualization titles are not displayed in the android app

Sharmyn28:
The current PR has:

- verify if the object/android-visualization (home, program, dataset) has a title. 
- in case it doesn't the name that comes from the API should be displayed in the table. 
- the name from the API should not be saved in the datastore.
- [DHIS2-12339](https://jira.dhis2.org/browse/DHIS2-12339)

_Title and name displayed_
image

_Values saved in datastore_
image


src/pages/Analytics/VisualizationQuery.js
```diff
@@ -0,0 +1,19 @@
...
+export const apiFetchVisualizations = async (dataEngine, visualization) => {
+    try {
+        const visualizationData = await dataEngine.query({
+            visualization: visualizationQuery(visualization),
+        })
+        return visualizationData.visualization.visualizations
+    } catch (error) {
+        console.log('Error: ', error)
+    }
+}
```
mediremi, line 19: Instead of using the data engine, would it be possible to use `useDataQuery` and have a `useVisualizations` hook instead? This would allow this app to benefit from [the caching functionality introduced in app-runtime v3](https://developers.dhis2.org/blog/2021/11/app-platform-v8#new-feature-usedataquery-caches-and-deduplicates-queries). It would also ensure that the loading and error cases are handled.

```js
const query = {
    resource: 'visualizations',
    params: ({ ids }) => ({
        fields: ['id', 'name', 'displayName'],
        filter: `id:in:[${ids}]`,
        paging: false,
    })
}

// see https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state
const usePrevious = value => {
  const ref = useRef()
  useEffect(() => {
    ref.current = value
  })
  return ref.current
}

export const useVisualizations = ids => {
  const prevIds = usePrevious(ids)
  const { called, loading, error, data, refetch } = useDataQuery(query, {
    lazy: true
  })

  useEffect(() => {
    if (!isEqual(ids, prevIds)) {
      refetch({ ids })
    }
  })
  
  return {
    loading: !called || loading,
    error,
    visualizations: data?.visualization.visualizations
  }
}
```

which can then be used like so:

```js
// src/pages/Analytics/Dataset/DatasetAnalyticList.js

const DatasetAnalyticList = ({
    visualizations,
    handleVisualizations,
    disable,
}) => {
    const { loading, error, visualizations: visualizationAPI } = useVisualizations(
        getVisualizationIdList(visualizations)
    )
    const { datasetList } = useReadDatasetQuery()
    const [rows, setRows] = useState()
    const [initialRows, setInitialRows] = useState()
    const [groups, setGroups] = useState()

    useEffect(() => {
        if (visualizations && datasetList) {
            const { visualizationsByDatasets, groupList } =
                prepareRows(visualizations, datasetList, visualizationAPI)
            setRows(visualizationsByDatasets)
            setGroups(groupList)
            setInitialRows(visualizationsByDatasets)
        }
    }, [visualizations, datasetList])

    useEffect(() => {
        if (rows && initialRows && !isEqual(rows, initialRows)) {
            handleVisualizations(rowsToDataStore(rows))
        }
    }, [rows])

    return (
        <>
            {loading && (
              <CircularLoader small />
            )}
            {error && (
              <NoticeBox error title={i18n.t('Error loading visualizations')}>
                {error.message}
              </NoticeBox>
            )}
            {!isEmpty(rows) && (
                <DatasetTable
                    rows={rows}
                    changeRows={setRows}
                    disabled={disable}
                />
            )}

            <NewDatasetVisualization
                disable={disable}
                visualization={rows}
                handleVisualization={setRows}
                groups={groups}
                handleGroups={setGroups}
            />
        </>
    )
}
```

mediremi (approved):

PR #131, opened 2022-04-20
fix: update PropTypes source

Sharmyn28:
This PR changes **PropTypes** source from `@dhis2/prop-types` to `prop-types`


HendrikThePendric:
I've had a look and don't think I have permission to do this (or at least I can't work out how I would do it). Austin had a suspicion it might be related to the node version, so I tried pinning that to 14, but that didn't help. We tend to have a `.github/workflows/dhis2-deploy-netlify.yml` file in our projects which holds the relevant config for deploying to Netlify, but it seems we don't have that here. TBH I'm a bit confused. Perhaps @amcgee could have a look?

HendrikThePendric:
I managed to get some elevated permissions in Netlify which allowed me to reset the build cache. It's still failing though. I'm not sure how this is happening exactly. Austin recommended adding the `D2_VERBOSE` environment variable, so that the build process would provide some more (and hopefully useful) output. But I am struggling to work out where I would add that for this repo (normally I would add it in `.github/workflows/dhis2-deploy-netlify.yml` but this project doesn't have that file)...

@Sharmyn28 Could you try to organise that, so we get a better handle on what's going on?

HendrikThePendric:
Update: I managed to add the `D2_VERBOSE` env variable in the Netlify dashboard now. There is more output in the deploy log, but nothing giving me a eureka moment....

Sharmyn28:
> I managed to get some elevated permissions in Netlify which allowed me to reset the build cache. It's still failing though. I'm not sure how this is happening exactly. Austin recommended adding the `D2_VERBOSE` environment variable, so that the build process would provide some more (and hopefully useful) output. But I am struggling to work out where I would add that for this repo (normally I would add it in `.github/workflows/dhis2-deploy-netlify.yml` but this project doesn't have that file)...
> 
> @Sharmyn28 Could you try to organise that, so we get a better handle on what's going on?

Hi @HendrikThePendric I found this [.yml file](https://github.com/dhis2/ui/blob/master/.github/workflows/dhis2-deploy-netlify.yml)  I'm copying it, let me know if there is something else I can do

HendrikThePendric (approved):

PR #134, opened 2022-04-25
fix: [program dataset] adjust component width

Sharmyn28:
This PR adjust the width of some Select components like program and dataset selection

image

### DIFF
--- a/src/pages/Appearance/Datasets/DialogDatasetSpecificSetting.js
+++ b/src/pages/Appearance/Datasets/DialogDatasetSpecificSetting.js
@@ -37,7 +37,7 @@ const DialogDatasetSpecificSetting = ({
                     ) : (
                         <Select
                             name="id"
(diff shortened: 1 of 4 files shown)


src/pages/Appearance/Datasets/DialogDatasetSpecificSetting.js
```diff
@@ -37,7 +37,7 @@ const DialogDatasetSpecificSetting = ({
                     ) : (
                         <Select
                             name="id"
-                            inputWidth="300px"
+                            inputWidth="350px"
```
simonadomnisoru, line 40: Hey @Sharmyn28 , 

Does it make sense to use the `dhis2/ui` `spacers` constant here? Something like this:
 `inputWidth={spacers.dp384}`. 

Thanks!

simonadomnisoru (approved):

PR #136, opened 2022-05-13
feat: edit titles in visualization Home Data set and Program [DHIS2-12737]

Sharmyn28:
This PR allows users to edit the visualization title in the Analytics section (Home, Program, Data set).

- Modal that only allows changing visualization title
- Modal shows visualizations properties like program/dataset and group

_Home Visualization_  [DHIS2-12737](https://jira.dhis2.org/browse/DHIS2-12737)
image

_Data set Visualization_  [DHIS2-12739](https://jira.dhis2.org/browse/DHIS2-12739)
image

_Program Visualization_  [DHIS2-12738](https://jira.dhis2.org/browse/DHIS2-12738)
image


tomzemp (commented):
Looks like it's working really well @Sharmyn28!

I had a question about the general approach for this refactor:
- wondering why you created three separate EditVisualization components for each of the pages where you need this modal? Would it not be easier to just write this code once and reuse it, so that maintaining/refactoring the modal is easier?
- similar question as to why you created a new NotEditableGroup component rather than just modifying the existing VisualizationGroup component to be disabled when used in the modal?

Sharmyn28:
> Looks like it's working really well @Sharmyn28!
> 
> I had a question about the general approach for this refactor:
> 
>     * wondering why you created three separate EditVisualization components for each of the pages where you need this modal? Would it not be easier to just write this code once and reuse it, so that maintaining/refactoring the modal is easier?
> 
>     * similar question as to why you created a new NotEditableGroup component rather than just modifying the existing VisualizationGroup component to be disabled when used in the modal?

Hi @tomzemp thanks for the comment, I made some changes to reuse the component EditVisualization and VisualizationGroup

src/components/analyticVisualization/EditVisualization.js
```diff
@@ -0,0 +1,96 @@
...
+    ModalTitle,
+} from '@dhis2/ui'
+import {NotEditableElement, VisualizationGroup, VisualizationTitle} from "./index";
+
+const EditVisualization = ({open, settings, handleChange, currentGroup, groups, handleEdit, handleClose, type}) => {
+    const [typeName, useTypeName] = useState('')
+    const [label, useLabel] = useState('')
+
+    const chooseTypeName = type => {
+        switch (type) {
```
simonadomnisoru, line 19: NIT: use a [lookup table pattern](https://omwri.medium.com/javascript-patterns-lookup-tables-26bbaf693e24) instead of the `switch` statement.

src/components/analyticVisualization/EditVisualization.js
```diff
@@ -0,0 +1,96 @@
...
+        chooseTypeName(type)
+    }, [type])
+
+    return (
+        <>
+            {open && (
+                <Modal position="middle">
+                    <ModalTitle>{i18n.t('Edit {{label}} visualization', {label})}</ModalTitle>
+                    <ModalContent>
+                        { type !=='home' &&
```
simonadomnisoru, line 44: Hey, 
I believe it is useful to move the string `home` and all the other `type` possible values to a `const` object. WDYT? 
Thanks!

src/components/analyticVisualization/VisualizationGroup/VisualizationGroup.js
```diff
@@ -26,10 +27,38 @@ const findDefaultGroup = (type, groupList, settings) => {
...
+        const groupFound = findGroup(type, { groupList, groupId }, settings)
+        if (disabled) {
+            if (!isDefaultGroup(groupFound)) {
+                setGroup(!!groupFound)
+                setTitle(groupFound.name)
+            } else {
+                setGroup(false)
+            }
+        }
+    }, [groupList])
```
simonadomnisoru, line 61: Hey,
I am curious to know if `type`, `groupId`, `disabled`, and `settings` can be part of the dependency array. When their values change the `useEffect` hook should run and compute again?
Thanks!

src/components/analyticVisualization/EditVisualization.js
```diff
@@ -0,0 +1,96 @@
...
+
+const EditVisualization = ({open, settings, handleChange, currentGroup, groups, handleEdit, handleClose, type}) => {
+    const [typeName, useTypeName] = useState('')
+    const [label, useLabel] = useState('')
+
+    const chooseTypeName = type => {
+        switch (type) {
+            case 'home':
+                useTypeName('')
+                useLabel('Home')
```
tomzemp, line 22: I think these strings need to be translated as they're passed through as the label of the NonEditableItem below?

src/components/analyticVisualization/EditVisualization.js
```diff
@@ -0,0 +1,96 @@
...
+                useTypeName('programName')
+                useLabel('Program')
+                break
+            case 'dataset':
+                useTypeName('datasetName')
+                useLabel('Data set')
+        }
+    }
+
+    useEffect(() => {
```
tomzemp, line 34: It looks like type is just a prop to this component and not changeable from this component or its children, so it seems like the useEffect should not be necessary?

src/components/analyticVisualization/VisualizationGroup/VisualizationGroup.js
```diff
@@ -26,10 +27,38 @@ const findDefaultGroup = (type, groupList, settings) => {
     return defaultGroup
 }
 
-export const VisualizationGroup = ({ settings, onChange, groupList, type }) => {
-    const { data: id } = useSystemId()
+const isDefaultGroup = group => group.name === 'default'
```
tomzemp, line 30: NIT: probably better to use `const isDefaultGroup = group => group.name === DEFAULT` as you've defined the constant already

src/components/analyticVisualization/EditVisualization.js
Sharmyn28: Hi, thanks for the suggestion. I'm changing it to:

```
const typeName = typeNameList[type] || ''
const label = labelList[type] || ''
```

src/components/analyticVisualization/EditVisualization.js
Sharmyn28: Thanks Tom. I'm adding this:

```
'home': i18n.t('Home'),
'program': i18n.t('Program'),
'dataset': i18n.t('Data set')
```

src/components/analyticVisualization/EditVisualization.js
Sharmyn28: Hi, 
currently, I'm only using the `type` home, but I can create a `const` object with `type` constants and the ones that come from the pattern you mentioned before. Like this:

```
const sections = {
    home: {
        type: 'home',
        typeName: '',
        label: i18n.t('Home'),
    },
    program: {
        type: 'program',
        typeName: 'programName',
        label: i18n.t('Program'),
    },
    dataset: {
        type: 'dataset',
        typeName: 'datasetName',
        label: i18n.t('Data set'),
    }
}
```

src/components/analyticVisualization/VisualizationGroup/VisualizationGroup.js
Sharmyn28: Hey, I'm using the `useEffect` hook to:
- refetch the id: the random id I get from that will be assigned if a new group has been created.
- In case I'm editing the visualization title: to validate if the visualization belongs to a group and assign the name/title of this group 

If I add `type`, `groupId`, `disabled`, and `settings` to the dependency array, then the id will be generated multiple times. I need only one new id each time. What do you think?

Sharmyn28: Maybe a `useEffect` with an empty dependency array because I only need one id each time.

simonadomnisoru (approved):

PR #138, opened 2022-05-14
fix: list hides visualizations with Today and Last 30 Days periods

Sharmyn28:
This PR fixes the visualization periods by changing the ids of Today/This Day and Last 30 Days.

[DHIS2-13220](https://jira.dhis2.org/browse/DHIS2-13220)

**Tested environment:**

[debug ](https://debug.dhis2.org/dev/)
_Visualization: TEST_AND id: EJEfci2t55b_
image

_Webapp: visualization is part of the list_
image

[previous1](https://play.dhis2.org/android-previous1)
_Visualization: Android: # of child - Today/last30Days id: EJEfci2t55b_
image

_Webapp: Android: # of child - Today/last30Days_
image


src/components/analyticVisualization/VisualizationSearch/periods.js
```diff
@@ -1,9 +1,9 @@
-export const TODAY = 'today' || 'thisDay'
+export const TODAY = 'today' && 'thisDay'
```
tomzemp, line 1: this will evaluate to 'thisDay', which I guess will work in the sense that `thisDay` is the relative period property on the visualizations list of relative periods, but if the goal is to get it to accept both `today` and `thisDay`, I think you would need to change the logic here to be something like

```
export const TODAY = 'today'
export const THIS_DAY = 'thisDay'
```

and then include both in your array of relative periods.

tomzemp (approved):

PR #139, opened 2022-06-13
feat: improve visualization list

Sharmyn28:
This PR improves the list of visualizations, showing all visualizations, and disabling visualizations that can't be used in the Android app.

[DHIS2-13381](https://jira.dhis2.org/browse/DHIS2-13381)

_Visualization list:_ 
image

image

### DIFF
--- a/src/components/analyticVisualization/VisualizationSearch/helper.js
+++ b/src/components/analyticVisualization/VisualizationSearch/helper.js
@@ -1,4 +1,5 @@
 import mapValues from 'lodash/mapValues'
+import orderBy from 'lodash/orderBy'
 import {
     LAST_12_MONTHS,
     LAST_12_WEEKS,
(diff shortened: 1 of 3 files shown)


src/components/analyticVisualization/VisualizationSearch/helper.js
```diff
@@ -97,3 +98,6 @@ const isValidOrgUnit = visualization =>
     visualization.userOrganisationUnit ||
     visualization.userOrganisationUnitChildren ||
     visualization.userOrganisationUnitGrandChildren
+
+export const orderVisualizations = visualizationList =>
+    orderBy(visualizationList, [item => item.valid === true], ['desc'])
```
tomzemp, line 103: I was going to suggest `orderBy(visualizationList, ['valid'], ['desc'])` but it looks like there are a lot of visualizations that have null value for valid (I don't know the app logic well enough to comment if that is appropriate or not.

tomzemp (approved):
lgtm!

I added a comment on visualizations missing valid property as I noticed it during review, but your approach works for dealing with that.

PR #142, opened 2022-07-08
ci: replace semantic commit checks with new workflow

Sharmyn28:


HendrikThePendric:
@ismay I've updated the repo settings so they now include the two commit linting actions as required checks.

PR #146, opened 2022-09-06
ci: run linting and tests on ci

ghost:
I've mocked the applayout component as it was getting too complex to render the entire app realistically. I think it'll be good to add react-testing lib and create more tests on a component by component basis.

I've also fixed any issues that could be fixed automatically (that's why the PR is so big unfortunately). All the other issues were eslint issues that I've disabled for now. I figure that it's more practical to at least have some linting than none, and then we can address the overridden rules when we can.

Most changes are autofixes and can be safely ignored (if you trust eslint and prettier), the only files with manual changes are:

- [app.test.js](https://github.com/dhis2/android-settings-app/pull/146/files#diff-1270de3bbb227435f61afc0333dccf3e1bcde4a082bf31a61acf489ae9bb3200)
- [package.json](https://github.com/dhis2/android-settings-app/pull/146/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)
- [verify app workflow](https://github.com/dhis2/android-settings-app/pull/146/files#diff-1d835fdd19fc1d6a0129eb898bf4593f444624610ad0ee1311c1a7ae906f70b7)
- [eslint config](https://github.com/dhis2/android-settings-app/pull/146/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5)


Sharmyn28:
What do we need to do if we move the app to that netlify workflow?

Sharmyn28:
> Ok, yeah it seems that resetting the cache on master did the trick: #147. Might be a false positive, but it seems likely at least. Don't know why I didn't think of this earlier.
> 
> So there's no rush in moving to our netlify workflow then. We can just tackle that when there's time, at a later stage.

I just approved the PR #147. I'll merge it (crossing fingers)

Sharmyn28 (approved):

PR #158, opened 2023-02-16
feat: use new tracker importer as default [DHIS2-14741]

Sharmyn28:
If the current API version is 2.40 or higher, the app should use the new tracker importer as the default option.

[DHIS2-14741](https://dhis2.atlassian.net/browse/DHIS2-14741)

- add a notice box/alert to inform the new tracker importer should be used as default
- use new tracker importer should be checked as a default option

_Notice box and default value_
image

_Datastore_
image


simonadomnisoru (approved):
Looks good!

PR #173, opened 2023-08-07
fix: allow set 0 reserved values [DHIS2-15652]

Sharmyn28:
This PR allows users to set the value 0 for reserved values. In case there is no value set, it uses the default value.

[DHIS2-15652](https://dhis2.atlassian.net/browse/DHIS2-15652)

image


simonadomnisoru (approved):
LGTM

PR #211, opened 2025-02-06
feat: add the quick actions table to the specific programs modal

Chisomchima:


src/pages/Appearance/Programs/QuickActionsSettings.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const QuickActionsSettings = ({ handleChange, settings }) => {
+    const [selectedActions, setSelectedActions] = useState([])
+    useEffect(() => {
+        if (settings?.quickActions) {
+            setSelectedActions(
+                settings.quickActions.map((action) => action.actionId)
+            )
+        }
+    }, [settings])
```
Birkbjo, line 23: Do we really need an `useEffect` here, wouldnt just setting the initialValue of `selectedActions` be enough? 
Will `settings.quickActions` ever update while this is rendered - and do we want that?

PR #219, opened 2025-03-10
feat: create custom intents menu and home page [DHIS2-18317]

Sharmyn28:
This PR creates custom intents section (no real data).

[DHIS2-18317](https://jira.dhis2.org/browse/DHIS2-18317)

- This PR it is not using real data
- Custom Intents menu section in the sidebar
- Custom Intents card in the overview page
- Custom Intents main page with components like:
  - footer buttons: save and reset
  - cards (name and description): list of custom intents 
  - notice box when there is no data
  - create button

_Menu section_
image

_Overview page_
image

_Custom Intent no data notice box_
image

_Custom Intent with cards_
image

[DHIS2-18317]: https://dhis2.atlassian.net/browse/DHIS2-18317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/Intents/CustomIntentsList.js
```diff
@@ -0,0 +1,52 @@
...
+            handleSettings(rows)
+        }
+    }, [rows])
+
+    return (
+        <>
+            {loading && <CircularLoader />}
+            {rows.length > 0 && <RowList rows={rows} disable={disable} />}
+
+            <NewIntent disable={disable} />
```
tomzemp, line 27: I think it would probably be easier to put this at CustomIntents.js level, so then this CustomIntentsList remains more strictly just a UI rendering of the current list?

src/pages/Intents/CustomIntentsList.js
```diff
@@ -0,0 +1,52 @@
...
+const CustomIntentsList = ({ settings, handleSettings, disable }) => {
+    const loading = false
+    const rows = settings
+    const initialRows = useMemo(() => {
+        return settings
+    }, [settings])
+
+    useEffect(() => {
+        if (rows && initialRows && !isEqual(rows, initialRows)) {
+            handleSettings(rows)
```
tomzemp, line 18: I know this is a placeholder for now, but I assume that for when the add/update/delete functionality edit gets added, we'd want to do it with functions passed down from CustomIntents.js and not having an effect firing based on change of state from here? (At the moment, `rows` also comes from the parent component, so there's also no actual component for the state to respond to anyway)

src/pages/Intents/CustomIntents.js
```diff
@@ -0,0 +1,80 @@
...
+    const initialValues = useMemo(() => {
+        return customIntents
+    }, [customIntentSettings])
+    const [mutate, { error, data }] = useDataMutation(saveCustomIntentsMutation)
+
+    useEffect(() => {
+        initialValues &&
+        customIntentSettings &&
+        !isEqual(initialValues, customIntentSettings)
+            ? setDisableSave(false)
```
tomzemp, line 33: I feel like this might just be possible as `const disabledSave = !isEqual(initialValues, customIntentSettings)` but that might depend on how much the function rerenders in the final implementation. 

Or, maybe even easier, you could keep track of initial values in a more-expanded-purpose `useCustomIntentSetttings` hook and then return a value that indicates whenever the "form state" is dirty/differs from the initial values (which would also help separate some of the UI/api logic more)

src/pages/Intents/CustomIntents.js
```diff
@@ -0,0 +1,80 @@
...
+const CustomIntents = () => {
+    const { hasAuthority } = useIsAuthorized()
+    const { customIntents } = useReadCustomIntentsDataStore()
+    const [disableSave, setDisableSave] = useState(true)
+    const [customIntentSettings, setCustomIntentSettings] =
+        useState(IntentsList)
+    const initialValues = useMemo(() => {
+        return customIntents
+    }, [customIntentSettings])
+    const [mutate, { error, data }] = useDataMutation(saveCustomIntentsMutation)
```
tomzemp, line 27: It might also make sense to just export the mutation from the a common hook (e.g. combine this and `useReadCustomIntentsDataStore` into something more like `useCustomIntents`)

src/pages/Intents/intentsDatastoreQuery.js
```diff
@@ -0,0 +1,76 @@
...
+                await mutateSharing({ id: metadata?.[CUSTOM_INTENTS]?.id })
+                setCustomIntents([])
+            } catch (error) {
+                setError(error)
+            } finally {
+                setLoading(false)
+            }
+        }
+
+        if (isEmpty(data) && !loadingNamespace) {
```
tomzemp, line 64: this will not fire when get to `dataStore/${NAMESPACE}/${CUSTOM_INTENTS}` fails because ${CUSTOM_INTENTS} key does not exist (I assume that is possible when you already have some android settings). In this case, you'd need to check (!loading || error), with the potential to check the actual error statusCode to see if it aligns with what you'd expect with the key not existing

src/pages/Intents/intentsDatastoreQuery.js
```diff
@@ -0,0 +1,76 @@
...
+    const [loading, setLoading] = useState(true)
+    const [error, setError] = useState(null)
+
+    useEffect(() => {
+        const fetchData = async () => {
+            try {
+                setLoading(true)
+                await mutate()
+                const metadata = await refetch()
+                await mutateSharing({ id: metadata?.[CUSTOM_INTENTS]?.id })
```
tomzemp, line 55: 👍 yay to setting the sharing settings!
Is the intent that only super users should have access to these resources, because this what seems will happen (when I tested with `admin` user, they could not update ${NAMESPACE} namepsace because they did not have permission).
I'm not sure if we also in that sense would also want to check the metadata sharing for a logged in user, so that we could export this to the CustomIntent component and disable save if they do not have write access?

src/pages/Intents/intentsDatastoreQuery.js
```diff
@@ -0,0 +1,76 @@
...
+    customIntents: {
+        resource: `dataStore/${NAMESPACE}/${CUSTOM_INTENTS}/metaData`,
+    },
+}
+
+export const useReadCustomIntentsDataStore = () => {
+    const { data, loading: loadingNamespace } = useDataQuery(
+        getCustomIntentsQuery
+    )
+    const [mutate] = useDataMutation(createKeyCustomIntent)
```
tomzemp, line 42: I feel like renaming the `mutate` and `refetch` here would be helpful for ease of readability, e.g. `mutate: createKeyCustomIntent`, refetch: setSharing`)

src/pages/Intents/intentsDatastoreQuery.js
```diff
@@ -0,0 +1,76 @@
...
+    const { refetch } = useDataQuery(getMetadataQuery, { lazy: true })
+    const [mutateSharing] = useDataMutation(updateSharingMutation)
+    const [customIntents, setCustomIntents] = useState([])
+    const [loading, setLoading] = useState(true)
+    const [error, setError] = useState(null)
+
+    useEffect(() => {
+        const fetchData = async () => {
```
tomzemp, line 50: I find the name of this function a bit confusing as it's not strictly fetching data but rather setting up customIntents key if it doesn't exist?

src/pages/Intents/intentsDatastoreQuery.js
```diff
@@ -0,0 +1,76 @@
...
+ * update data store
+ * key: Custom intents
+ * */
+
+export const saveCustomIntentsMutation = {
+    resource: `dataStore/${NAMESPACE}/${CUSTOM_INTENTS}`,
+    type: 'update',
+    data: ({ settings }) => ({
+        ...settings,
+        lastUpdated: new Date().toJSON(),
```
tomzemp, line 17: this should work in the sense of getting ISO/UTC time, but I guess is inconsistent with the general practice of DHIS2 of recording times as server time zones, e.g. if my server is in Oslo, my update for now (13 March 14:42), would be recorded as `2025-03-13T14:42:00.000`, rather than as `2025-03-13T13:42:00.000Z`

tomzemp (approved):
looks good! 🎉

I left a few comments, but a bunch of them might be more relevant when you start implementing functionality for actually adding/editing/deleting intents.

A couple of other points I noticed:
- in the cards for the custom intents, the description comes from `systemSettings` property, but I didn't see that on the sample payload in the [specs](https://dhis2.atlassian.net/wiki/spaces/MOB/pages/804257793/Reqmt.+Custom+Intents).
- the UI looks like it conforms to the specs, but I'm a bit surprised that the `Delete` button is not styled as `destructive` or `destructive secondary`
- the save modal refers to the "DataStore", but elsewhere in the app it's written as two words ("data store").

src/pages/Intents/intentsDatastoreQuery.js
Sharmyn28: Only users with `ALL` or `M_androidsettingsapp` authority have permission to create/update. The `hasAuthority` hook checks if the user has the authority :smile:

PR #222, opened 2025-04-03
feat: global completion spinner default value [DHIS2-19398]

Sharmyn28:
This PR adds a default value to global completion spinner.

[DHIS2-19398](https://jira.dhis2.org/browse/DHIS2-19398)

Default value: `false` (unchecked)

[DHIS2-19398]: https://dhis2.atlassian.net/browse/DHIS2-19398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/Appearance/Programs/__tests__/createInitialGlobalSpinner.test.js
+++ b/src/pages/Appearance/Programs/__tests__/createInitialGlobalSpinner.test.js
@@ -1,0 +1,58 @@
+import { createInitialGlobalSpinner } from '../helper'
+
+test('Generate default values', () => {
+    const result = {
+        completionSpinner: false,
+        disableReferrals: false,
+        disableCollapsibleSections: false,
+    }
+
+    expect(createInitialGlobalSpinner({})).toMatchObject(result)
+})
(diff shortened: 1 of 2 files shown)


tomzemp (approved):
Is this subtly different that what you had before? Like it seems that if `prevDetails = {completionSpinner:null}` then `{completionSpinner=false}= prevDetails` would result in completionSpinner:null now? Whereas before it would have resulted in completionSpinner:false?

But maybe this is also not an issue based on the implementation?

Sharmyn28:
> Is this subtly different that what you had before? Like it seems that if `prevDetails = {completionSpinner:null}` then `{completionSpinner=false}= prevDetails` would result in completionSpinner:null now? Whereas before it would have resulted in completionSpinner:false?
> 
> But maybe this is also not an issue based on the implementation?

Thanks for bringing the `null` case, I'm going back to the previous function and adding a few tests.

PR #227, opened 2025-05-12
feat: create custom intent

Chisomchima:


src/pages/Intents/IntentIdentifiers.js
```diff
@@ -0,0 +1,135 @@
...
+]
+
+const IntentIdentifiers = ({
+    formData,
+    onChange,
+    dataElements,
+    attributes,
+}) => {
+    const [selectedElementType, setSelectedElementType] = useState('')
+    const [list, setList] = useState([])
```
tomzemp, line 31: I don't think you need to keep these in state. You could just conditionally use dataElements or attributes based on the value of `selectedElementType`?

src/components/field/MultiSelectField.js
```diff
@@ -0,0 +1,46 @@
...
+    ...props
+}) => (
+    <Field
+        label={label}
+        required={required}
+        error={error}
+        validationText={error ? validationText : ''}
+    >
+        <MultiSelect
+            placeholder="Choose an option"
```
tomzemp, line 22: Needs to be wrapped in i18n.t

src/pages/Intents/IntentIdentifiers.js
```diff
@@ -0,0 +1,135 @@
...
+                        label: item.displayName,
+                        value: item.id,
+                    }))}
+                    inputWidth="300px"
+                    required
+                />
+            )}
+
+            <div style={{ marginBottom: '20px', marginTop: '20px' }}>
+                <MultiSelectField
```
tomzemp, line 109: is there a reason not just to use the dhis2-ui's <MultiSelectField> component ?

src/pages/Intents/ResponseForm.js
```diff
@@ -0,0 +1,42 @@
...
+    const handleChange = (field, value) => {
+        const updated = { ...responseData, [field]: value }
+        setResponseData(updated)
+        onChange('response.data', updated)
+    }
+
+    return (
+        <div style={{ display: 'grid', gap: '8px' }}>
+            <TextField
+                label="Argument"
```
tomzemp, line 21: these need to be translatable with i18n.t

src/pages/Intents/ResponseForm.js
```diff
@@ -0,0 +1,42 @@
...
+    const [responseData, setResponseData] = useState(data)
+
+    useEffect(() => {
+        setResponseData(data)
+    }, [data])
+
+    const handleChange = (field, value) => {
+        const updated = { ...responseData, [field]: value }
+        setResponseData(updated)
+        onChange('response.data', updated)
```
tomzemp, line 15: why not just fully make use of the setByPath function you added and update only the specific field e.g update onChange('response.data.argument',value), rather than updating the whole data object and then updating that?

tomzemp (commented):
Looks great 👏! Thanks for putting all of this together.

_Thing that I think we need to update:_
- for the request.arguments portion of the payload, I believe that the intent is for the key to be dynamic. E.g. so right now, when you add `key: animal`, `value: bear`, we create `{"request":{"arguments":[{"key": "animal", "value": "bear"}]}}`, but I believe from the specs the intent is that we create a dynamic key/value pair, e.g. we do `{"request":{"arguments":{"animal":"bear"}}}`

_Other recurring minor comments:_
- I'd move the styling to css modules, as we do not typically do inline styling
- there are several strings missing i18n.t wrapping. I've specifically commented most of them, but I've not commented on ones that were close by.

Other than that, I've put some smaller, one-off comments in the review.

PR #228, opened 2025-05-20
feat: add and edit custom intent dev

Chisomchima:


src/pages/Intents/CustomIntentsList.js
```diff
@@ -0,0 +1,170 @@
...
+
+    const handleSave = (newIntent) => {
+        const updated = [...settings]
+        if (newIntent.uid) {
+            const idx = updated.findIndex((s) => s.uid === newIntent.uid)
+            if (idx > -1) {
+                updated[idx] = newIntent
+            }
+        } else {
+            updated.push({ ...newIntent, uid: generateDhis2Id() })
```
Sharmyn28, line 57: To keep the consistency we should use `id` rather than `uid`

src/pages/Intents/DialogCustomIntents.js
```diff
@@ -0,0 +1,200 @@
...
+                                    {i18n.t('Response')}
+                                </Tab>
+                            </TabBar>
+                        </div>
+                        <div className={styles.contentWrapper}>
+                            {renderStep()}
+                        </div>
+                        <ButtonStrip end>
+                            {step > 0 && (
+                                <Button onClick={prevStep}>
```
Sharmyn28, line 155: `Prev` and `Next` button should be small :smile:

tomzemp:
@Sharmyn28 @Chisomchima : while making the PR updates, I also noticed 
- that you could not delete a custom intent when there was only one (this was due to changes I made on a useEffect to get all items in dependency array; basically, when it got to no items, it mistakenly set to the initial values) (I've updated logic now, so this doesn't happen)
- we were not making posts to create customIntents endpoint when it was missing (it was created on initial save of a custom intent, but then we missed the sharing modification) (updated this as well)

PR #229, opened 2025-05-20
feat: custom intents edit [DHIS2-19107]

tomzemp:
This is PR for https://dhis2.atlassian.net/browse/DHIS2-19107

This PR makes some minor edits to connect the custom intents list with the hook for fetching custom intents.

In addition, I've made some general clean up (e.g. to add i18n, align the text of the fields closer to what is specified in the [specs](https://dhis2.atlassian.net/wiki/spaces/MOB/pages/804257793/Reqmt.+Custom+Intents).)


src/pages/Intents/DialogCustomIntents.js
```diff
@@ -63,7 +64,7 @@ const DialogCustomIntents = ({
             arguments:
                 specificSettings?.request?.arguments?.length > 0
                     ? specificSettings.request.arguments
-                    : [{ key: '', value: '' }],
+                    : [{ key: '', value: '', id: generateDhis2Id() }],
```
tomzemp, line 67: @Chisomchima: I added this id because the first was missing an id, but I'm not sure if we want to use id here in the first place? It's not in the specs, so we could also just make the updates based on index of the {key,value} item in the array and skip adding an id in the first place?

src/pages/Intents/IntentIdentifiers.js
```diff
@@ -81,38 +76,36 @@ const IntentIdentifiers = ({
...
 
             <TextField
                 name="description"
-                label={i18n.t('Description')}
+                label={i18n.t('Intent description')}
                 onChange={(e) => onChange('description', e.value)}
                 value={formData.description || ''}
             />
 
-            {!edit && (
```
tomzemp, line 97: I didn't think this should be conditionally hidden on edit, but maybe I missed something when looking at the specs?

Chisomchima (approved):
Thanks Tom, lgtm!

